PR #36, opened 2015-03-01
Update libjsonnet.h

loverszhaokai:
I wonder whether it is a mistake ?


sparkprime:
Those are doxygen constructs, although I've never actually run it through doxygen.  There are some more in state.h as well.

loverszhaokai:
Get it.

PR #37, opened 2015-03-01
No args case

loverszhaokai:
Print usage() when ./jsonnet without args


sparkprime:
So the current documented behavior is for this case is the filename defaults to stdin.  That may not be particularly useful but it is documented as doing that :)

loverszhaokai:
OK, it maybe better handled in the future.

sparkprime:
https://github.com/google/jsonnet/commit/41c11d09a7d7c7b4c07a144946ab3672494721dd

PR #68, opened 2015-09-12
Add argument to -m flag to specify the output directory for multi file output

### LINKED ISSUE #66: jsonnet flag for setting output directory for -m
Currently, `jsonnet -m` writes the output files in the current directory. It would be useful to have a way to set the output directory, either as an optional argument to `-m` or as a separate flag.

davidzchen:
Fixes #66

### DIFF
--- a/BUILD
+++ b/BUILD
@@ -14,6 +14,8 @@ genrule(
           "echo >> $@",
 )
 
+# TODO(dzc): Remove the includes = ["."] lines from all cc_* targets once
+# bazelbuild/bazel#445 is fixed.
 cc_library(
     name = "jsonnet-common",
     srcs = [
@@ -30,6 +32,7 @@ cc_library(
         "static_error.h",
         "vm.h",
     ],
+    includes = ["."],
 )
 
 cc_library(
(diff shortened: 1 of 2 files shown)


sparkprime:
Is it possible to re-use -o for this?

sparkprime:
Actually I guess the list of files output should go to -o

davidzchen:
I am not sure whether using `-o` for this is better. In general, using `-o` when outputting a single file allows one to rename the output file to something other than the default. However, for multiple file output, the basenames of the files cannot be changed, and the only thing that can be easily changed is which directory the files are written to. I know that adding yet another flag is not ideal, but I think simply setting a directory for multiple file output is easier to do than listing all the output files on `-o`.

Another option would be to have `-m` take the output directory an optional argument. What do you think?

By the way, I am not sure what is causing the CI failures. I am unable to reproduce locally using `make test`.

sparkprime:
I am OK with changing -m

davidzchen:
Done.

davidzchen:
Resolved merge conflicts.

PR #75, opened 2015-09-30
Restructure source tree.

davidzchen:
* Move all C++ sources for core library into core/
* Move source for Jsonnet command line too into cmd/
* Move sources for Python bindings into python/
* Move Jsonnet standard library into stdlib/

To better illustrate the new source layout, see the `restructure` branch in my fork: https://github.com/davidzchen/jsonnet/tree/restructure

The main motivation for this is to more cleanly organize the core library and the different language bindings. This way, the Java bindings would be added to a java/ directory.


sparkprime:
LGTM

PR #76, opened 2015-10-20
Add desugaring.{cpp,h} to BUILD.

davidzchen:
Related to #3 

The new `core/desugaring.{cpp,h}` was added to the Makefile but not to BUILD. :)


sparkprime:
Thanks

PR #77, opened 2015-10-20
Add C++ API.

### LINKED ISSUE #71: C++ API
While Jsonnet is written in C++, the core API it exposes is a C API. It would be nice to have a proper C++ API as well.

davidzchen:
Fixes #71

## Background

I started working on this to get a better idea of what the Java bindings should look like. This implementation is a thin wrapper on top of the C API, providing some C++ niceties such as `std::string`, STL containers, and `std::function` for import callbacks.

I am opening this pull request as an RFC for now to get feedback on the API's design. There are still some additional changes I would like to add before this is fully ready, including:

* Adding a test for import callbacks
* Fixing Makefile. Currently, it no longer builds the `.so` files by default for some reason, even though they are under `$(ALL)`
* Add documentation for the C++ import callbacks.


Makefile
```diff
@@ -48,16 +48,25 @@ LIB_SRC = \
...
+LIB_CPP_SRC = \
+	$(LIB_SRC) \
+	cpp/jsonnet.cc
+
+LIB_CPP_OBJ = $(LIB_CPP_SRC:.cpp=.o)
+
 ALL = \
 	jsonnet \
 	libjsonnet.so \
+	libjsonnet-cpp.so \
```
davidzchen: For some reason, running `make` only builds `jsonnet` but not the `.so` libraries, even though the `so`'s are under `ALL`.

cpp/jsonnet.cc
```diff
@@ -0,0 +1,132 @@
...
+    const char *json = c2;
+    while (*c2 != '\0') ++c2;
+    ++c2;
+    c = c2;
+    outputs->insert(std::make_pair(filename, json));
+  }
+}
+}  // namespace
+
+bool Jsonnet::EvaluateFileMulti(
```
davidzchen: The error condition is not handled correctly for the multi APIs since I am returning a `map<string, string>` for the output parameter and thus the error message now gets parsed into a key in the map. I see two options:

1. Have a separate parameter for returning the error message
2. Stash the error message in a class member and expose a `GetLastError` method for returning the error message.

What do you think?

cpp/jsonnet.h
```diff
@@ -0,0 +1,161 @@
...
+#include <cstring>
+#include <map>
+
+extern "C" {
+  #include "core/libjsonnet.h"
+}
+
+namespace jsonnet {
+
+template <typename C>
```
davidzchen: I need to add some documentation and an example for this.

Makefile
davidzchen: This is expected. Running `make all` builds the `.so`'s.

davidzchen:
This PR is ready for review. I have just added some unit tests for the C++ API.

Since the `jsonnet_jpaths_add` API has been added, and the primary use case of the import callback has been to add directories to the import path, I am leaving out the import callback from this API for now.

Currently, the C++ API is a simple wrapper around the C APIs. This is not ideal since Jsonnet itself is implemented in C++. It would be better to refactor common logic out of libjsonnet.c and expose both C and C++ APIs. This way, we would able to also expose classes such as [FmtOpts](https://github.com/google/jsonnet/blob/master/core/formatter.h) in the C++ API, which would be more suitable than having separate methods for setting each formatter option.

BUILD
```diff
@@ -8,7 +8,7 @@ filegroup(
 genrule(
     name = "gen-std-jsonnet-h",
     srcs = ["stdlib/std.jsonnet"],
-    outs = ["core/std.jsonnet.h"],
+    outs = ["std.jsonnet.h"],
```
sparkprime: Seems unrelated to this PR :)

BUILD
```diff
@@ -38,17 +39,24 @@ cc_library(
         "core/string_utils.h",
         "core/unicode.h",
         "core/vm.h",
-        "include/libjsonnet.h",
+    ],
+    deps = [
+        "//include:libjsonnet",
```
sparkprime: Apologies that I haven't RTFM but how does this work?  Does it take the libjsonnet cc_library rule and use the 
only the "includes" param from it?

WORKSPACE
```diff
@@ -0,0 +1,16 @@
+new_http_archive(
```
sparkprime: What's all this? :)

BUILD
sparkprime: ...nevermind I just found include/BUILD

cpp/libjsonnet++.cc
```diff
@@ -0,0 +1,157 @@
+// Copyright 2015 Google Inc. All rights reserved.
```
sparkprime: Would this make sense as a header only library?

BUILD
davidzchen: I was getting an Bazel build error about being unable to find the generated header file:

```
❯❯❯ bazel build //:jsonnet-common
INFO: Found 1 target...
ERROR: /Volumes/Ocean/Projects/google/jsonnet/BUILD:17:1: undeclared inclusion(s) in rule '//:jsonnet-common':
this rule is missing dependency declarations for the following files included by 'core/desugarer.cpp':
  'bazel-out/local_darwin-fastbuild/genfiles/std.jsonnet.h'.
Target //:jsonnet-common failed to build
Use --verbose_failures to see the command lines of failed build steps.
INFO: Elapsed time: 1.457s, Critical Path: 1.29s
```

This might be a bug in the C/C++ rules related to using the `includes` attribute while looking for a header file in the genfiles tree. Changing the genrule to output the file in the current package rather than `core/` worked.

BUILD
davidzchen: No worries. Yeah, `//include:libjsonnet` refers to the `:libjsonnet` target in `include/`.

WORKSPACE
davidzchen: This downloads the gmock archive so that we can use it for the unit tests. This is similar to CMake's [ExternalProject](https://cmake.org/cmake/help/v3.0/module/ExternalProject.html) mechanism but friendlier. :)

Here's the documentation for [`new_http_archive`](http://bazel.io/docs/be/workspace.html#new_http_archive).

cpp/libjsonnet++.cc
davidzchen: What do you mean? Are you referring to the `cc_library` rules under `include/`? The problem is that since the target for `cpp/libjsonnet++.cc` are in `cpp/BUILD`, directly adding `../include/libjsonnet++.h` would cross package boundaries, and Bazel would complain. The way to include headers in a separate package is to add a `cc_library` rule for the header and then add that rule to `deps`.

sparkprime: I noticed it did not include many new headers, so if it (libjsonnet++) were a header only library, it would not need a new build target and would be much simpler.  No need to do it in this PR but it's worth considering.

BUILD
sparkprime: Ok as long as it works, no need for it to be the same location as the makefile for example.

sparkprime:
lgtm

cpp/libjsonnet++.cc
davidzchen: I see what you mean. Makes sense. Also, if we end up refactoring libjsonnet.cpp and exposing both a C and C++ APIs, then that will essentially be the case.

PR #84, opened 2015-12-02
Fix the python build (again?)

benley:
desugarer.o looks like a typo.  desugaring.o exists and seems to make things work again.


sparkprime:
Thanks for the fix

PR #85, opened 2015-12-08
Add a simplistic "make install" target

benley:
To really do this properly would probably mean setting up autoconf
and/or automake, which doesn't sound like a ton of fun.  This mimics a
subset of what such a system would produce, which should make life
easier for distribution packagers.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -37,6 +37,17 @@ LDFLAGS ?=
 
 SHARED_LDFLAGS ?= -shared
 
+# Prefix for all installed files
+prefix = /usr/local
+
+exec_prefix = ${prefix}
+bindir = ${exec_prefix}/bin
+includedir = ${prefix}/include
+libdir = ${exec_prefix}/lib
+DESTDIR =
+INSTALL = install -c
+INSTALL_DATA = ${INSTALL} -m 644
+
 ################################################################################
 # End of user-servicable parts
 ################################################################################
@@ -70,8 +81,10 @@ ALL_HEADERS = \
 	core/std.jsonnet.h \
 	include/libjsonnet.h
 
+.PHONY: default
 default: jsonnet
 
+.PHONY: all
 all: $(ALL)
 
 TEST_SNIPPET = "std.assertEqual(({ x: 1, y: self.x } { x: 2 }).y, 2)"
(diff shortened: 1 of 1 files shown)


sparkprime:
Does the autoconf stuff you're working on need this or replace this?

benley:
It would replace this.

benley:
I have the autotools stuff working now, but haven't worked out how to get the tests back into it yet.  Almost.

davidzchen:
Just wondering, what is the motivation for choosing Autotools over CMake?

benley:
To be completely honest I didn't give that a lot of consideration; it's just because I've used autotools a bit in the past and it seems like the universal lowest-common-denominator build system.  Personally I'd prefer to use Bazel, but having a generic `./configure;make;make install` build method would be good for user adoption and distro packaging and such.

benley:
libtool does make the autotools build pretty slick though; it correctly builds shared libraries with pkgconfig metadata on linux, macos, and probably a bunch of other platforms without further fiddling.

sparkprime:
cmake does build visual studio files which might be useful.  However last time I tried to build Jsonnet on msvc, it crashed the compiler...  So probably best for people to use g++ on windows for now.

Ditto on bazel -- it's nice but I think we'll need the fallback and then we'll have to maintain both and that would be a pain.

I have fixed the integration tests btw, merge master to get that.

sparkprime:
Just remembered that we do actually have a bazel BUILD file now... :)

benley:
I'll get back to the autotools project in a few days, I think.  Might also try doing it with cmake to see how that goes.

sparkprime:
The irony is Jsonnet is a great tool for describing builds :)

benley:
Heh, in that case [Repobuild](https://github.com/chrisvana/repobuild) might be an appealing build system!  (I'm not being serious, but you may find it amusing: it's like bazel, but uses JSON instead of a python-like dsl)

davidzchen:
From Repobuild's README:

> Similar to Google's BUILD file system of old (gconfig + make)

Now that's a bit of old history. :)

Similarly, Kythe also had its own custom [Campfire](https://github.com/google/kythe/blob/75b9fc4ec6e85bd0c30c89ae9474444003aa62f9/buildtools/CAMPFIRE) build system, which was also similar to BUILD but used JSON, before they fully switched to Bazel.

PR #91, opened 2016-01-08
BUILD: fix build now that sandboxing is enabled.

mikedanese:
Now that sandboxing is enabled in newer versions of bazel,
we need to declare all inputs or the build will fail.

http://bazel.io/blog/2015/09/11/sandboxing.html


sparkprime:
Thanks.  It'd be good to have a regression test for building with bazel.

sparkprime:
And setup.py for that matter...

PR #95, opened 2016-01-08
stdlib: implement json merge

mikedanese:
Adds a merge function to the stdlib as described in:

https://tools.ietf.org/html/rfc7386
https://tools.ietf.org/html/draft-ietf-appsawg-json-merge-patch-07

There's some wierdness around patching nulls. I couldn't figure out how to remove a node from the json object so some of the test cases have an extra null in the expected output which seems okay. I've left a comment where are expected differs from the spec's expected differs from our expected.

### DIFF
--- a/stdlib/std.jsonnet
+++ b/stdlib/std.jsonnet
@@ -840,6 +840,30 @@ limitations under the License.
                     aux(a, b, i, j + 1, acc) tailstrict;
         aux(a, b, 0, 0, []) tailstrict,
 
+    mergePatch(target, patch)::
+        if std.type(patch) == "object" then
+            local target_object =
+                if std.type(target) == "object" then target else {};
+
+            local target_fields =
+                if std.type(target_object) == "object" then std.objectFields(target_object) else [];
+
+            local null_fields = [k for k in std.objectFields(patch) if patch[k] == null];
+            local both_fields = std.setUnion(target_fields, std.objectFields(patch));
+
+            {
+                [k]:
+                    if !std.objectHas(patch, k) then
+                        target_object[k]
+                    else if !std.objectHas(target_object, k) then
+                        std.mergePatch(null, patch[k]) tailstrict
+                    else
+                        std.mergePatch(target_object[k], patch[k]) tailstrict
+                for k in std.setDiff(both_fields, null_fields)
+            }
+        else
+            patch,
+
     objectFields(o)::
         std.objectFieldsEx(o, false),
 
--- a/test_suite/error.equality_function.jsonnet.golden
+++ b/test_suite/error.equality_function.jsonnet.golden
@@ -1,3 +1,3 @@
 RUNTIME ERROR: Cannot test equality of functions
-	std.jsonnet:889:17-41	function <anonymous>
+	std.jsonnet:913:17-41	function <anonymous>
 	error.equality_function.jsonnet:17:1-31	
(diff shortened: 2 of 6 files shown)


sparkprime:
Try this on for size:

```
    merge(target, patch)::
        if std.type(patch) == "object" then 

            local target_fields =
                if std.type(target) == "object" then std.objectFields(target) else [];

            local null_fields = [k for k in std.objectFields(patch) if patch[k] == null];
            local both_fields = std.setUnion(target_fields, std.objectFields(patch));
            {
                [k]:
                    if !std.objectHas(patch, k) then
                        target[k]
                    else if !std.objectHas(target, k) then
                        patch[k]
                    else
                        util.merge(target[k], patch[k])
                for k in std.setDiff(both_fields, null_fields)
            }
        else
            patch
```

sparkprime:
Also, the reason you have some failing test cases is because you've changed std.jsonnet and the line numbers in the "golden" stacktraces for unrelated tests no-longer align.  To fix that, use refresh_golden.sh to rewrite the "golden" files with the new content (and sanity check with git diff).

sparkprime:
Maybe this should be called mergePatch, and it really calls out for the diff equivalent as well, although not necessarily in this PR.

sparkprime:
It looks like 7396 is the better RFC as it obsoletes 7386

mikedanese:
It looks like the pseudocode algorithm for applying a patch did not change between the three RFC revisions so the difference must be in the http parts.

Nice! this seems to work:
```
    merge(target, patch)::
        if std.type(patch) == "object" then 
            local target_object =
                if std.type(target) == "object" then target else {};

            local target_fields =
                if std.type(target_object) == "object" then std.objectFields(target_object) else [];

            local null_fields = [k for k in std.objectFields(patch) if patch[k] == null];
            local both_fields = std.setUnion(target_fields, std.objectFields(patch));

            {
                [k]:
                    if !std.objectHas(patch, k) then
                        target_object[k]
                    else if !std.objectHas(target_object, k) then
                        patch[k]
                    else
                        std.merge(target_object[k], patch[k]) tailstrict
                for k in std.setDiff(both_fields, null_fields)
            }
        else
            patch,
```
It fixes all extra nulls except for the one in the last test case.

sparkprime:
I think it will work with this as the inner block:

```
                    if !std.objectHas(patch, k) then
                        target[k]
                    else if !std.objectHas(target, k) then
                        std.merge(null, patch[k])
                    else
                        std.merge(target[k], patch[k])
```

mikedanese:
I think that is equivalent since std.merge(null, patch[k]) evaluates to patch[k] (always hits the else). The problem is with the else return patch branch. To conform to the spec we would have to walk the patch object and strip out all explictly null fields.

sparkprime:
Nah std.merge(null, patch[k]) is recursive into patch[k]

std.merge(x, null) evaluates to null though

mikedanese:
Ok I gotcha. I've updated the patch.

sparkprime:
Looks like error.equality_function.jsonnet.golden needs updating

test_suite/merge.jsonnet.golden
```diff
@@ -0,0 +1 @@
+true
```
sparkprime: Absence of a .golden file means it should expect it to output just true, so you can delete this file.

mikedanese:
@sparkprime addressed comments and build is all green

Issue #101, opened 2016-02-06
Support array slicing notation


mikedanese:
It would be very useful to support python style array slicing:
```
arr[start:end:stride]
```
I think this can be performed by desugaring to makeArray

mikedanese:
I added parsing support here:

https://github.com/google/jsonnet/commit/c6e3ad6fc228331b3ad47c1b38b87da813203f8c

I'm deciding whether this should be implemented in the vm or desugarring

PR #106, opened 2016-02-17
implement array slicing

mikedanese:
Incomplete pass at #101

```console
$ cat exp.jsonnet 
local arr = std.range(0, 10);
arr[5:7]
$ ./jsonnet exp.jsonnet
[
   5,
   6
]
```

### DIFF
--- a/core/ast.h
+++ b/core/ast.h
@@ -309,11 +309,17 @@ struct Importstr : public AST {
  * One of index and id will be nullptr before desugaring.  After desugaring id will be nullptr.
  */
 struct Index : public AST {
+    bool isSlice;
     AST *target;
     AST *index;
+    AST *end;
+    AST *step;
     const Identifier *id;
     Index(const LocationRange &lr, AST *target, AST *index, const Identifier *id)
-      : AST(lr, AST_INDEX), target(target), index(index), id(id)
+      : AST(lr, AST_INDEX), isSlice(false), target(target), index(index), end(nullptr), step(nullptr), id(id)
+    { }
+    Index(const LocationRange &lr, AST *target, AST *index, AST *end, AST *step)
+      : AST(lr, AST_INDEX), isSlice(true), target(target), index(index), end(end), step(step), id(nullptr)
     { }
 };
 
--- a/core/desugaring.cpp
+++ b/core/desugaring.cpp
@@ -21,7 +21,7 @@ limitations under the License.
 #include "parser.h"
 
 static LocationRange E;  // Empty.
-    
+
 static unsigned long max_builtin = 24;
 BuiltinDecl jsonnet_builtin_decl(unsigned long builtin)
 {
@@ -438,7 +438,7 @@ class Desugarer {
                     } break;
                 }
             }
-                    
+
             ast_ = in;
 
         } else if (auto *ast = dynamic_cast<Assert*>(ast_)) {
@@ -511,13 +511,44 @@ class Desugarer {
             // Nothing to do.
 
         } else if (auto *ast = dynamic_cast<Index*>(ast_)) {
-            desugar(ast->target, obj_level);
-            if (ast->id != nullptr) {
-                assert(ast->index == nullptr);
-                ast->index = str(ast->id->name);
-                ast->id = nullptr;
+            if (ast->isSlice) {
+                if (ast->index == nullptr)
+                    ast->index = make<LiteralNull>(ast->location);
+                desugar(ast->index, obj_level);
+
+                if (ast->end == nullptr)
+                    ast->end = make<LiteralNull>(ast->location);
+                desugar(ast->end, obj_level);
+
+                if (ast->step == nullptr)
+                    ast->step = make<LiteralNull>(ast->location);
+                desugar(ast->step, obj_level);
+
+                ast_ = make<Apply>(
+                    ast->location,
+                    make<Index>(
+                        E,
+                        std(),
+                        str(U"slice"),
+                        nullptr),
+                    std::vector<AST*>{
+                        ast->target,
+                        ast->index,
+                        ast->end,
+                        ast->step,
+                    },
+                    false,
+                    false
+                );
+            } else {
+                desugar(ast->target, obj_level);
+                if (ast->id != nullptr) {
+                    assert(ast->index == nullptr);
+                    ast->index = str(ast->id->name);
+                    ast->id = nullptr;
+                }
+                desugar(ast->index, obj_level);
             }
-            desugar(ast->index, obj_level);
 
         } else if (auto *ast = dynamic_cast<Local*>(ast_)) {
             for (auto &bind: ast->binds)
--- a/core/parser.cpp
+++ b/core/parser.cpp
@@ -526,6 +526,10 @@ namespace {
             return tok;
         }
 
+        void push(Token tok) {
+            tokens.push_front(tok);
+        }
+
         Token peek(void)
         {
             Token tok = tokens.front();
@@ -955,7 +959,7 @@ namespace {
                         }
                     } while (true);
                     return alloc->make<Array>(span(tok, next), elements, got_comma);
-                    
+
                 }
 
                 case Token::PAREN_L: {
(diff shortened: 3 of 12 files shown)


mikedanese:
Just pushed support for `[:4]` and `[4:]` notations

```console
$  jsonnet git:(slice) ✗ cat exp.jsonnet 
local arr = std.range(0, 10);
{
    foo: arr[5:7],
    bar: arr[:7],
    baz: arr[5:],
}
$  jsonnet git:(slice) ✗ ./jsonnet exp.jsonnet 
{
   "bar": [
      0,
      1,
      2,
      3,
      4,
      5,
      6
   ],
   "baz": [
      5,
      6,
      7,
      8,
      9,
      10
   ],
   "foo": [
      5,
      6
   ]
}
```

mikedanese:
This doesn't support step parameter yet.

core/ast.h
```diff
@@ -317,6 +318,15 @@ struct Index : public AST {
     { }
 };
 
+struct Slice : public AST {
```
sparkprime: Is there an advantage to having this as a separate AST if it can only be used inside the [ ] of an Index AST?

sparkprime: Basically you could extend Index to add the last and step ASTs which are null if they're not provided.

core/parser.cpp
```diff
@@ -1186,9 +1186,42 @@ namespace {
 
                     Token op = pop();
                     if (op.kind == Token::BRACKET_L) {
-                        AST *index = parse(MAX_PRECEDENCE);
+                        const int max = 2;
+                        bool isSlice = true;
```
sparkprime: The coding convention here is that locals on the stack follow the C style of foo_bar whereas fields of classes use fooBar.  While I'm at it, macros and constants are FOO_BAR.

core/parser.cpp
```diff
@@ -1186,9 +1186,42 @@ namespace {
...
+                        AST *first = nullptr;
+                        AST *last = nullptr;
+
+                        for (int i = 0; i < max; i++) {
```
sparkprime: I think this code would be simplified by unrolling the loop and folding the constant i values in.  You could handle the breaks by nesting ifs.

test_suite/error.slice_out_of_bounds.sugar.jsonnet
```diff
@@ -0,0 +1,3 @@
+local arr = std.range(0, 5);
+
+arr[0:10]
```
sparkprime: We should exactly replicate the semantics of Python

>>> range(0, 5)[0:10]
[0, 1, 2, 3, 4]

Also supporting negative numbers.  This can all be implemented in std.slice() :)

test_suite/slice.sugar.jsonnet
```diff
@@ -0,0 +1,37 @@
...
+        input:  arr[a:b],
+        output: [ 2, 3 ],
+    },
+    {
+        input:  arr[a:b+1],
+        output: [ 2, 3, 4 ],
+    },
+];
+
+std.foldl(
```
sparkprime: It'd be an interesting project to design a unit testing framework for Jsonnet.

sparkprime:
Good work!  Sorry it took me a few days to get around to reviewing it.

stdlib/std.jsonnet
```diff
@@ -88,6 +88,9 @@ limitations under the License.
     range(from, to)::
         std.makeArray(to - from + 1, function(i) i + from),
 
+    slice(arr, from, to)::
+        std.makeArray(to - from, function(i) arr[i + from]),
```
sparkprime: This should also work on strings and fail explicitly if arr is not a string or array type.  There is already str.substr but its semantics aren't right (see comment below).  So we should deprecate str.substr() and use this instead.

core/parser.cpp
sparkprime: Ideally the code should read like the BNF:

```
e ::= ...
    | e [ e [ ':' [ [ e ] [ ':' [ e ] ] ] ] ]
```
I think that's the right syntax anyway :)

test_suite/error.slice_out_of_bounds.sugar.jsonnet
mikedanese: Python slicing seems crazy complex: 

https://github.com/rfk/pypy/blob/master/pypy/objspace/std/listobject.py#L1362-L1382

I'll look into duplicating the behavior exactly into std.slice()

core/ast.h
mikedanese: Done

core/parser.cpp
mikedanese: Done.

core/parser.cpp
mikedanese: Unrolled.

test_suite/slice.sugar.jsonnet
mikedanese: too bad JUnit is taken :)

mikedanese:
Ok, I need to implement the python behavior in std.slice now.

mikedanese:
I've implemented python behavior for all positive value (index, end, step) inputs. I think it's reasonable to support negative inputs later on.

test_suite/slice.sugar.jsonnet
```diff
@@ -0,0 +1,106 @@
...
+        output: arr,
+    },
+    {
+        input:  arr[:1000],
+        output: arr,
+    },
+    //{
+    //    input:  arr[::],
+    //    output: arr,
+    //},
```
mikedanese: @sparkprime This is breaking with:
```
FAIL (exit code): slice.sugar.jsonnet
This run's output:
STATIC ERROR: slice.sugar.jsonnet:32:21-22: Not a unary operator: ::
```
I'll investigate what is going on but it's not immediately obvious to me.

stdlib/std.jsonnet
```diff
@@ -88,6 +88,35 @@ limitations under the License.
...
+            local invar =
+                // loop invariant
+                {
+                    indexable: indexable,
+                    end: end,
+                    step: step,
+                    length: std.length(indexable),
+                    type: std.type(indexable)
+                };
+            local build =
```
sparkprime: local build(slice, cur) =

test_suite/slice.sugar.jsonnet
sparkprime: Oh this is bad unfortunately -- { x:: 3 } is lexed as the single token ::

sparkprime: I guess you can change the parser to check for that, a bit like C++ does with C<vector<int>>

core/desugaring.cpp
```diff
@@ -511,13 +511,47 @@ class Desugarer {
...
-            desugar(ast->target, obj_level);
-            if (ast->id != nullptr) {
-                assert(ast->index == nullptr);
-                ast->index = str(ast->id->name);
-                ast->id = nullptr;
+            if (ast->isSlice) {
+                if (ast->end != nullptr) {
+                    desugar(ast->end, obj_level);
+                } else {
+                    ast->end = length(ast->target);
```
sparkprime: If you desugared them to null, you could test for null in std.slice and implement the full semantics that way

sparkprime: >>> [1,2,3][None:None:None]
[1, 2, 3]

In fact this is mandatory for python compatibility :)

test_suite/slice.sugar.jsonnet
sparkprime: A workaroudn is also to do [ : : ]

mikedanese: Ok, fixed with https://github.com/mikedanese/jsonnet/blob/slice/core/parser.cpp#L1201-L1208 and enabled the test.

mikedanese:
@sparkprime addressed all review comments.

test_suite/slice.sugar.jsonnet
mikedanese: I can do it without push() but it requires a whole nother branch which feels gross

stdlib/std.jsonnet
```diff
@@ -88,6 +88,41 @@ limitations under the License.
...
+                if cur >= invar.end || cur >= invar.length then
+                    slice
+                else
+                    build(
+                        if invar.type == "string" then
+                            slice + invar.indexable[cur]
+                        else
+                            slice + [ invar.indexable[cur] ],
+                        cur + invar.step
+                    ) tailstrict;
```
mikedanese: O(n^2) :( but would rather optimize this in the vm for array concatinations where one side is not used again.

```console
$ jsonnet git:(slice): cat exp.jsonnet 
std.range(0, std.extVar("vars").count)[2:100000000000:2]
$ jsonnet git:(slice): time ./jsonnet --code-var vars='{"count": 1000}' exp.jsonnet > /dev/null 
./jsonnet --code-var vars='{"count": 1000}' exp.jsonnet > /dev/null  0.25s user 0.00s system 99% cpu 0.258 total
$ jsonnet git:(slice): time ./jsonnet --code-var vars='{"count": 10000}' exp.jsonnet > /dev/null                                                                                             
./jsonnet --code-var vars='{"count": 10000}' exp.jsonnet > /dev/null  1.98s user 0.01s system 99% cpu 1.990 total
$ jsonnet git:(slice): time ./jsonnet --code-var vars='{"count": 100000}' exp.jsonnet > /dev/null
./jsonnet --code-var vars='{"count": 100000}' exp.jsonnet > /dev/null  46.23s user 0.30s system 99% cpu 46.594 total
```

sparkprime: Yes I am not sure how to solve this.  There are various instances of it throughout the code.

The Python approach is to make std.join O(n) and then do everything in terms of that.

Currently std.join is O(n^2) and making it O(n) would mean making it native.

There may be a better way though.

mikedanese: I'm curious about implementing the internal representation with a persistent data structure. Copy on write minus the deep copy.

Issue #115, opened 2016-02-29
jsonnet -e '[1,2,3][1::]' not a binary operator: "::"


mikedanese:
Fix this once the fmt is merged.

PR #116, opened 2016-03-01
break :: into :, : anywhere in a slice

mikedanese:
#115

### DIFF
--- a/test_suite/slice.sugar.jsonnet
+++ b/test_suite/slice.sugar.jsonnet
@@ -49,6 +49,14 @@ local arrCases = [
         output: arr,
     },
     {
+        input: arr[1::],
+        output: [1, 2, 3, 4, 5],
+    },
+    {
+        input: arr[1::2],
+        output: [1, 3, 5],
+    },
+    {
         input: arr[::2],
         output: [0, 2, 4],
     },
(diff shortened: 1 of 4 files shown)


test_suite/slice.sugar.jsonnet
```diff
@@ -49,6 +49,14 @@ local arrCases = [
         output: arr,
     },
     {
+        input: arr[1::],
```
mikedanese: It's not possible to run the formatter on these tests as it denatures the test.

sparkprime: Yeah there are several cases of that already, they should have a .fmt.golden file to hold the output of formatting those files (because they do not format to themselves).  If I accidentally broke the test, please fix it ;)

sparkprime: You'll want to run ./refresh_fmt_golden.sh on this one.

mikedanese: Ah ok. Updated .fmt.golden. It looks like you made some changes but manually reverted anything that would have modified the tests so lgtm.
 https://github.com/google/jsonnet/commit/17c59acbfd455e42ab6bae0e5517157be4bf0b90#diff-5dd5fae04796372f66a075da6ff11ffc

PR #132, opened 2016-03-18
Add unit test for libjsonnet C API. 

davidzchen:
Move Bazel build targets into BUILD files in respective directories rather than having a BUILD file in the root.


sparkprime:
LGTM

PR #133, opened 2016-03-18
Add unit tests for lexer

davidzchen:
This change adds unit tests for the lexer.

These tests were transferred from the lexer tests in https://github.com/google/go-jsonnet/blob/master/lexer_test.go.

These changes are made on top of #132, which should be merged first.

### DIFF
--- a/core/static_error.h
+++ b/core/static_error.h
@@ -17,6 +17,9 @@ limitations under the License.
 #ifndef JSONNET_STATIC_ERROR_H
 #define JSONNET_STATIC_ERROR_H
 
+#include <iostream>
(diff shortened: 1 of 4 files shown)


core/static_error.h
```diff
@@ -91,14 +94,21 @@ struct StaticError {
       : location(location), msg(msg)
     {
     }
+
+    std::string ToString() const
```
sparkprime: toString is used elsewhere

sparkprime:
Otherwise LGTM

davidzchen:
Rebased and addressed review comments.

PR #138, opened 2016-03-23
Drop namespace indents.

davidzchen:
As discussed offline, this change drops the indent within namespaces. This allows us to save
on horizontal whitespace and makes the code a bit easier to follow.

### DIFF
--- a/core/state.h
+++ b/core/state.h
@@ -19,415 +19,415 @@ limitations under the License.
 
 namespace {
 
-    /** Mark & sweep: advanced by 1 each GC cycle.
-     */
-    typedef unsigned char GarbageCollectionMark;
+/** Mark & sweep: advanced by 1 each GC cycle.
+ */
+typedef unsigned char GarbageCollectionMark;
 
-    /** Supertype of everything that is allocated on the heap.
-     */
-    struct HeapEntity {
-        GarbageCollectionMark mark;
-        virtual ~HeapEntity() { }
+/** Supertype of everything that is allocated on the heap.
+ */
+struct HeapEntity {
+    GarbageCollectionMark mark;
+    virtual ~HeapEntity() { }
+};
+
+/** Tagged union of all values.
+ *
+ * Primitives (<= 8 bytes) are copied by value.  Otherwise a pointer to a HeapEntity is used.
+ */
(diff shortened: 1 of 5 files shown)


mikedanese:
have we discussed standardizing on a .clang-format before?

davidzchen:
I don't think we have, but I also think it is a good idea to standardize on an established style guide and `.clang-format`, especially since we are getting more contributors. I think we should consider adopting the Google C++ Style Guide to both be consistent with most Google open source projects.

sparkprime:
Nope, the Google C++ style guide is terrible.  However I am definitely in favor of codifying the existing style and fixing any inconsistencies within.

davidzchen:
Why do you say that it is terrible?

core/state.h
```diff
@@ -19,415 +19,415 @@ limitations under the License.
...
-         *
-         * Note, this is non-const because we have to add cyclic references to it.
-         */
-        BindingFrame upValues;
+    /** The result when the thunk was forced, if filled == true. */
+    Value content;
 
-        /** The captured self variable, or nullptr if there was none.  \see CallFrame. */
-        HeapObject *self;
+    /** Used in error tracebacks. */
```
sparkprime: Looks like Github diffing is broken here?

davidzchen: Yes, looks that way. This patch is purely whitespace changes only aside from the additions of the `// namespace` comments to the closing braces for the namespaces.

sparkprime:
LGTM

sparkprime:
A good rant on the style guide here: https://www.linkedin.com/pulse/20140503193653-3046051-why-google-style-guide-for-c-is-a-deal-breaker

I cut my teeth on Ogre graphics engine and Bullet physics engine, so the style I use is similar to theirs.  It's also similar to the code I wrote at IBM for the X10 project.

davidzchen:
Yes, I have seen the rant, and I do not 100% agree with it. For example, it mentions that disallowing multiple inheritance is 100% a deal breaker, but I am not convinced that multiple inheritance is a "strength" in practice except for multiple interface inheritance.

In any case, let's move this discussion to a separate bug. I've opened #139 to discuss this.

PR #144, opened 2016-03-27
Adjust sanity checks now there is no BUILD in the root

sparkprime:
@davidzchen


davidzchen:
LGTM. Thanks for fixing this!

Once we have the Bazel build running in Travis, these issues will be easier to catch.

PR #154, opened 2016-04-02
Remove unnecessary check

sparkprime:
Variables should always be available in the stack.


sparkprime:
Should have used my @google.com address...

PR #164, opened 2016-04-25
Better functions

### LINKED ISSUE #147: Richer functions and more principled top-level parameterization
Context
=======

Jsonnet's original design used object inheritance (mixins) as the primary abstraction mechanism.  Since then, users have indicated that in many cases they prefer abstracting via functions.  Therefore, this proposal adds several features to make Jsonnet functions more productive.  These new features are intended to be conceptually and syntactically compatible with Python.  Additionally, an alternative to std.extVar() is proposed which allows top-level parameterization of configurations without creating the equivalent of a global variable.

1) Add default arguments
=========================

Function parameters can have default values.  To the function body, such parameters are no different to regular parameters.

```
$ jsonnet -e '(function(a, b, c=3) a + b + c)(1, 2)'
6
```

The default expression is bound in the outer scope.  E.g.:

```
$ jsonnet -e 'local a = 3; (function(a, b, c=a) c)(1, 2)'
6
```

2) Named arguments
==================

Function call syntax is extended to allow calling a function with named arguments.  All positional arguments must be given before all named arguments.

```
f(b=2, a=1)
f(1, b=2, c=3)
```

If no parameter exists by that name, an error is raised:

```
f(1, b=2, d=3)
RUNTIME ERROR: No such parameter "d"
```

3) Top-level functions
======================

Previously a Jsonnet file evaluating to a function would be rejected:

```
$ jsonnet -e 'function() 42'
RUNTIME ERROR: Couldn't manifest function in JSON output.
```

In this proposal such a function would be called with no arguments.

```
$ jsonnet -e 'function(a=2) 21 * a'
42
```

This only works if the function is callable with no arguments:

```
$ jsonnet -e 'function(a) 21 * a'
RUNTIME ERROR: Expected 1 argument, got 0.
```

And, only one level of function expansion is performed:

```
$ jsonnet -e 'function(a=2) function() 21 * a'
RUNTIME ERROR: Couldn't manifest function in JSON output.
```

To provide an alternative to std.extVar() that is not globally scoped (i.e., requires top-level parameters to be threaded through a large config), new CLI flags are added to provide arguments to top-level functions.

```
$ jsonnet -e 'function(a, b, c=3) a + b + c' --tla-code a=1 --tla-code b=2
6
```

Top-level functions can therefore have parameters without default arguments, and if no argument is provided on the CLI then an error is raised:

```
$ jsonnet -e 'function(a, b, c=3) a + b + c' --tla-code a=1
RUNTIME ERROR: Expected 2-3 arguments, got 1.
```

4) CLI flag cleanup
===================

The CLI flags have grown organically and are now somewhat messy.  Here is a the current state:

```
+----------------+--------------+----------------+-----------------+
| Flag           | Kind of data | Source of data | Accessible via  |
+----------------+--------------+----------------+-----------------+
| -V --var       | string       | commandline    | std.extVar(...) |
| -E --env       | string       | environment    | std.extVar(...) |
| -F --file      | string       | file           | std.extVar(...) |
|    --code-var  | jsonnet      | commandline    | std.extVar(...) |
|    --code-env  | jsonnet      | environment    | std.extVar(...) |
|    --code-file | jsonnet      | file           | std.extVar(...) |
+----------------+--------------+----------------+-----------------+
```

The meaning of the flags is not clearly indicated by their names.  Thus, the long forms of these flags is renamed along with the addition of a new dimension to the space (top-level func vs extvar).  The result is longer but more organized:

```
+--------------------+--------------+----------------+-----------------+
| Flag               | Kind of data | Source of data | Accessible via  |
+--------------------+--------------+----------------+-----------------+
| -V --ext-str       | string       | commandline    | std.extVar(...) |
| -E --ext-str-env   | string       | environment    | std.extVar(...) |
| -F --ext-str-file  | string       | file           | std.extVar(...) |
|    --ext-code      | jsonnet      | commandline    | std.extVar(...) |
|    --ext-code-env  | jsonnet      | environment    | std.extVar(...) |
|    --ext-code-file | jsonnet      | file           | std.extVar(...) |
| -T --tla-str       | string       | commandline    | Top-level func  |
|    --tla-str-env   | string       | environment    | Top-level func  |
|    --tla-str-file  | string       | file           | Top-level func  |
|    --tla-code      | jsonnet      | commandline    | Top-level func  |
|    --tla-code-env  | jsonnet      | environment    | Top-level func  |
|    --tla-code-file | jsonnet      | file           | Top-level func  |
+--------------------+--------------+----------------+-----------------+
```

sparkprime:
Fix #147

### DIFF
--- a/cmd/jsonnet.cpp
+++ b/cmd/jsonnet.cpp
@@ -79,12 +79,6 @@ void usage(std::ostream &o)
     o << "  -h / --help             This message\n";
     o << "  -e / --exec             Treat filename as code\n";
     o << "  -J / --jpath <dir>      Specify an additional library search dir\n";
-    o << "  -V / --var <var>=<val>  Specify an 'external' var to the given value\n";
-    o << "  -E / --env <var>        Bring in an environment var as an 'external' var\n";
-    o << "  -F / --file <var=<val>  Specify and 'external' var as the contents of a given file\n";
-    o << "  --code-var <var>=<val>  As --var but value is Jsonnet code\n";
-    o << "  --code-env <var>        As --env but env var contains Jsonnet code\n";
-    o << "  --code-file <var>=<val> As --file but file contents is Jsonnet code\n";
     o << "  -o / --output-file <file> Write to the output file rather than stdout\n";
     o << "  -m / --multi <dir>      Write multiple files to the directory, list files on stdout\n";
     o << "  -y / --yaml-stream      Write output as a YAML stream of JSON documents\n";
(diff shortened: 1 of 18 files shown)


mikedanese:
Can't wait to try this out. It looks like tla needs to be plumbed through the python api

sparkprime:
I'll work on that now, should be pretty trivial

PR #189, opened 2016-05-20
fix bazel build

### LINKED ISSUE #188: Bazel build is broken
```
% bazel build cmd:jsonnet
INFO: Found 1 target...
ERROR: /Users/shimin/jsonnet/core/BUILD:32:1: C++ compilation of rule '//core:parser' failed: osx_cc_wrapper.sh failed: error executing command external/bazel_tools/tools/cpp/osx_cc_wrapper.sh '-D_FORTIFY_SOURCE=1' -fstack-protector -fcolor-diagnostics -Wall -Wthread-safety -Wself-assign -fno-omit-frame-pointer '-std=c++0x' -iquote . -iquote ... (remaining 29 argument(s) skipped): com.google.devtools.build.lib.shell.BadExitStatusException: Process exited with status 1.
In file included from core/parser.cpp:29:
In file included from core/desugarer.h:24:
core/vm.h:21:10: fatal error: 'libjsonnet.h' file not found
#include "libjsonnet.h"
         ^
1 error generated.
Target //cmd:jsonnet failed to build
Use --verbose_failures to see the command lines of failed build steps.
INFO: Elapsed time: 1.412s, Critical Path: 1.24s
```
core/vm.h is in the //core package whereas libjsonnet.h is in the //include package.

guoshimin:
Fixes #188

### DIFF
--- a/cpp/testdata/BUILD
+++ b/cpp/testdata/BUILD
@@ -9,8 +9,8 @@ jsonnet_to_json_test(
     name = "example_test",
     src = "example.jsonnet",
     golden = "example_golden.json",
-    jsonnet = "//:jsonnet",
-    std = "//:std",
+    jsonnet = "//cmd:jsonnet",
+    std = "//stdlib",
 )
 
 jsonnet_to_json_test(
@@ -18,8 +18,8 @@ jsonnet_to_json_test(
     src = "importing.jsonnet",
     golden = "importing_golden.json",
     imports = ["."],
-    jsonnet = "//:jsonnet",
-    std = "//:std",
+    jsonnet = "//cmd:jsonnet",
+    std = "//stdlib",
 )
 
 jsonnet_to_json_test(
@@ -27,8 +27,8 @@ jsonnet_to_json_test(
     src = "invalid.jsonnet",
     golden = "invalid.out",
     error = 1,
-    jsonnet = "//:jsonnet",
-    std = "//:std",
+    jsonnet = "//cmd:jsonnet",
+    std = "//stdlib",
 )
 
 filegroup(
--- a/stdlib/BUILD
+++ b/stdlib/BUILD
@@ -1,8 +1,8 @@
 package(default_visibility = ["//visibility:public"])
 
 filegroup(
-  name = "stdlib",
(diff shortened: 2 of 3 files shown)


sparkprime:
@davidzchen This looks OK to me?  What do you think?

cpp/testdata/BUILD
```diff
@@ -9,26 +9,26 @@ jsonnet_to_json_test(
...
 
 jsonnet_to_json_test(
     name = "invalid_test",
     src = "invalid.jsonnet",
     golden = "invalid.out",
     error = 1,
-    jsonnet = "//:jsonnet",
-    std = "//:std",
+    jsonnet = "//cmd:jsonnet",
+    std = "//stdlib",
```
davidzchen: Come to think of it, I think we can remove the `std` attribute since the stdlib is baked into the jsonnet binary. @sparkprime is this correct?

stdlib/BUILD
```diff
@@ -1,5 +1,10 @@
 package(default_visibility = ["//visibility:public"])
 
+filegroup(
+  name = "stdlib",
```
davidzchen: Nit(BUILD convention): Indent 4 spaces

davidzchen:
Just have a few comments, otherwise LGTM.

cpp/testdata/BUILD
guoshimin: You may be right that it's unnecessary, but the jsonnet_to_json_test rule defined in the bazel repo has a default value for std, which is simply wrong for these targets.

stdlib/BUILD
guoshimin: Done.

guoshimin:
Comments addressed. PTAL.

cpp/testdata/BUILD
sparkprime: I think David's right but we can fix it some other time

sparkprime:
LGTM thanks

cpp/testdata/BUILD
davidzchen: Agreed. I will remove the `std` attribute from the Bazel Jsonnet rules in a separate change.

PR #191, opened 2016-05-24
avoid trailing spaces in manifestJsonEx output

guoshimin:
This PR avoids trailing spaces in manifestJsonEx output when the input value is an empty array or object. Apart from the aesthetic benefit, one major benefit is making the output eligible for the literal block syntax in yaml, which forbids trailing spaces.


sparkprime:
Nice job

PR #218, opened 2016-07-16
Fixed libjsonnet.h href in bindings.html

Neeke:
The href "http://github.com/google/jsonnet/blob/master/libjsonnet.h" is removed to "http://github.com/google/jsonnet/blob/master/include/libjsonnet.h".
So fixed it.


Neeke:
I signed it!

sparkprime:
Thanks for the fix

PR #224, opened 2016-07-26
Fix pagination in the 3rd page of the case study

dastergon:
The page enumeration was broken in the third page because there was a leading slash missing in the hyperlinks, and they raised 404 errors when trying to move backwards.


sparkprime:
Thanks

PR #242, opened 2016-10-05
Fix warnings in formatter.cpp on travis osx

sparkprime:


sparkprime:
Didn't fix the spurious failing osx build but merging anyway...

PR #258, opened 2016-11-17
Bazel rules for the python module

benley:
The python_repository repo rule digs up the local system's Python
headers and builds _jsonnet.so with them. Any downstream users of this
repo could easily supply a different python location or version as
needed.

I also rearranged the Python tests so they could be hooked in with both
the setup.py test command and a bazel py_test rule.


sparkprime:
Thanks

PR #260, opened 2016-11-18
add extVarOrNull and extVarOrEmpty

lvlv:
Hi Jsonnet folks,

I'm trying to add two functions: extVarOrNull and extVarOrEmpty, which returns Null or Empty when the var is not defined in ext-code or ext-str.

Here is my use case:

{
  local a = 4,
  local b = std.extVarOrNull("a"),
  "c": if b == null then a else b
}

$ jsonnet a.jsonnet 
{
   "c": 4
}

$ jsonnet a.jsonnet --ext-code "a=5"
{
   "c": 5
}

### DIFF
--- a/core/desugarer.cpp
+++ b/core/desugarer.cpp
@@ -31,7 +31,7 @@ struct BuiltinDecl {
     std::vector<String> params;
 };
 
-static unsigned long max_builtin = 25;
+static unsigned long max_builtin = 26;
 BuiltinDecl jsonnet_builtin_decl(unsigned long builtin)
 {
     switch (builtin) {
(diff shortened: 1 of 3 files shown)


sparkprime:
We deliberately didn't include this functionality, as since the extVar creates a global variable, it can cause surprising misbehaviors when code deep in the config depends on an extVar that isn't given.  If you want a default variable, please use the top-level-argument feature, which allows you to do defaults (which is safe since the variables have to be threaded deep into the config):

```
function(c = 4) {
"c": c
}
```

```
$ jsonnet a.jsonnet
{
   "c": 4
}
```

```
$ jsonnet a.jsonnet --tla-code c=5
{
   "c": 5
}
```

lvlv:
Rocks!

PR #265, opened 2016-11-25
Implement verbatim string literals

### LINKED ISSUE #246: raw/unescaped string literals?
We sometimes need to write regexes in jsonnet documents, and the double escaping can get a bit awkward, e.g.

``` python
"^\\nx\\.y\\.z\\.com\\n$"
```

A raw string literal syntax would be very nice, perhaps like python's:

``` python
r"^\nx\.y\.z\.com\n$"
```

Desired json output for the above, of course: `"^\\nx\\.y\\.z\\.com\\n$"`

benley:
Fixes #246

### DIFF
--- a/core/formatter.cpp
+++ b/core/formatter.cpp
@@ -427,7 +427,7 @@ class Unparser {
                 if (bind.functionSugar) {
                     unparseParams(bind.parenLeftFodder, bind.params, bind.trailingComma,
                                   bind.parenRightFodder);
-                } 
+                }
                 fill(bind.opFodder, true, true);
                 o << "=";
                 unparse(bind.body, true);
@@ -1324,7 +1324,7 @@ class FixIndentation {
             Indent new_indent = strong_indent
                                 ? newIndentStrong(first_fodder, indent, new_column)
                                 : newIndent(first_fodder, indent, new_column);
-                
+
             first = true;
             for (auto &element : ast->elements) {
                 if (!first) column++;
@@ -1452,7 +1452,7 @@ class FixIndentation {
                 if (bind.functionSugar) {
                     params(bind.parenLeftFodder, bind.params, bind.trailingComma,
                            bind.parenRightFodder, new_indent);
-                } 
+                }
                 fill(bind.opFodder, true, true, new_indent.lineUp);
                 column++;  // '='
                 Indent new_indent2 = newIndent(open_fodder(bind.body), new_indent, column + 1);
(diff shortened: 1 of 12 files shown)


sparkprime:
:D  Reviewing this is #1 priority when I'm back to work on Monday

core/formatter.cpp
```diff
@@ -464,6 +464,30 @@ class Unparser {
...
+                    if (*cp == U'"') {
+                        o << "\"\"";
+                    } else {
+                        std::string utf8;
+                        encode_utf8(*cp, utf8);
+                        o << utf8;
+                    }
+                }
+                o << "\"";
+            } else if (ast->tokenKind == LiteralString::VERBATIM_SINGLE) {
```
sparkprime: There's one other case to handle these, and that's in the FixIndentation pass.  It's exactly the same, except you increment the column counter rather than writing to the the stream.  So you'll still have to loop over ast->value and either add 2 or 1 depending no whether it's a quote.

core/lexer.cpp
```diff
@@ -507,6 +507,38 @@ Tokens jsonnet_lex(const std::string &filename, const char *input)
...
+                    throw StaticError(filename, begin, ss.str());
+                }
+                const char quot = *c;
+                c++;  // Advance beyond the opening quote.
+                for (; ; ++c) {
+                    if (*c == '\0')  {
+                        throw StaticError(filename, begin, "Unterminated verbatim string");
+                    }
+                    if (*c == quot) {
+                        if (*(c+1) == quot) {
```
sparkprime: Can you add a comment explaining that the '' and "" quoting is interpreted here, unlike the non-verbatim strings where it is done later by jsonnet_string_unescape.  This is OK in this case because no information is lost by resolving the repeated quote into a single quote, so we can go back to the original form in the formatter.

doc/language/spec.html
```diff
@@ -123,13 +123,17 @@ <h2 id="lexing">Lexing</h2>
...
 
-<li><i>string</i>: Which can have three quoting forms:
+<li><i>string</i>: Which can have five quoting forms:
 
 <ul>
 <li>Double-quoted, beginning with <code>"</code> and ending with the first subsequent non-quoted
 <code>"</code> </li>
 <li>Single-quoted, beginning with <code>'</code> and ending with the first subsequent non-quoted
 <code>'</code> </li>
+<li>Double-quoted verbatim, beginning with <code>@"</code> and ending with the first subsequent
```
sparkprime: You should probably say "non-quoted" again here, although the quoting scheme is different, it's still a quoting scheme.

benley:
I haven't forgotten about this - just haven't had time to return to it.  I'll try to finish it up soon.

sparkprime:
No problem, thanks for the update

core/lexer.cpp
benley: Done; pretty much used your verbiage verbatim.  (ha! verbatim! ... I'll show myself out)

doc/language/spec.html
benley: Done.

core/formatter.cpp
benley: Done, I think.

sparkprime:
Awesome... I already have a mental workitem to update the tutorial with some of the new features so I'll add this at the same time.

PR #268, opened 2016-12-05
Fix #267 argparams

### LINKED ISSUE #267: spec is wrong regarding argparams
In the spec http://jsonnet.org/language/spec.html both function application and function definition reference `argparams`:
- expr ( [ argparams ] )
- function ( [ argparams ] ) expr

where `argparams` is defined as
```
::=	id { , id } { , id = expr } [ , ]
|	id = expr { , id = expr } [ , ]

```
This is wrong because in function application, it should be
```
expr { , expr } { , id = expr } [ , ]
```

I see that the spec is also in github. I'm happy to submit a patch, but I'm not sure if the page is autogenerated from some other source.

sparkprime:
@guoshimin Think this addresses it

### DIFF
--- a/doc/language/spec.html
+++ b/doc/language/spec.html
@@ -253,7 +253,7 @@ set are currently parseable, the rest being reserved for possible future use.</p
 </td></tr> <tr><td></td><td> | </td><td>
 <i>expr</i>
 <code>(</code>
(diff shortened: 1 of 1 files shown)


guoshimin:
Thanks!

PR #279, opened 2017-01-11
Fix #266

### LINKED ISSUE #266: Reformatter introduces huge indent
```
$ cat test.jsonnet
0 +
1

$ jsonnet fmt -n 1 test.jsonnet
0 +
                                                                                                                                                                                                                                                                1
```

brbbs:
#266 is caused by `eval()` using `column` before it's initialized. On my machine, this ended up with `column` being set to 256. I didn't exhaustively check all the branches in `eval()`, but most seem to increment column before using it, so this issue only showed up for the `Binary` AST branch.

I'm not sure what the best way to add a test for this in the test_suite directory since we need to invoke jsonnet with the `-n` flag. Do you have any ideas?

### DIFF
--- a/test_suite/arith_float.jsonnet.fmt.golden
+++ b/test_suite/arith_float.jsonnet.fmt.golden
@@ -107,7 +107,7 @@ std.assertEqual(5.5 & 3.3, 1) &&
 std.assertEqual(4.5 | 1.9, 5) &&
 
 std.assertEqual(~4.5, -  #
-5) &&
+    5) &&
 
 std.assertEqual(~4.5, -5) &&
 std.assertEqual(-4.5, 0 - 4.5) &&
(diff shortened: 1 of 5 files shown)


sparkprime:
Nice!

I think we should just run the tests with -n 4.  I don't think that will break any of them.

sparkprime:
Ok it breaks 9 of them :)  It looks like some of them expose actual bugs in the formatter, so it's non-trivial to just turn it on.

sparkprime:
OK when I fix the other 2 open issues I'll be able to push a change to use -n 4 on all the fmt tests...  Might happen tonight but more likely tomorrow.

sparkprime:
Now it should be possible to add a test for this either as part of formatter.jsonnet or a separate test is fine too.  I actually already merged a change that initializes the column counter properly so your change will disappear if you rebase.  However the credit is still yours :)

PR #288, opened 2017-01-13
Add fmt test for toplevel binary expression

brbbs:
Github closed #279 when I rebased, but here's a test for you. I made it a separate test since the bug in #266 would only show up if the first AST processed was a Binary AST; the other ASTs would have initialized/modified the column.


sparkprime:
Thanks!

PR #291, opened 2017-01-26
Add stdlib md5 function

huggsboson:


huggsboson:
The md5 code is from this page:
http://www.zedwood.com/article/cpp-md5-function

PR #294, opened 2017-01-29
Fix formatter behavior around + and ()

### LINKED ISSUE #278: fmt creates less-than-ideal indentation
Hi, 
I ran into several cases where the auto formatter does not display the code in a reasonable ways. 

Here are some examples.:

```
// 1)
{
    objects():: [object]
                + if condition then [object2]
    else [],
}

// 2)
{
    objects():: [object]
                + if condition then
        [object2]
    else [],
}

// 3)
{
    objects():: [object] + if condition then
        [object2]
    else [],
}
```

I am using the the "-i -n 4 " options to fmt via vim-jsonnet. 
Can be seen here 
https://github.com/google/vim-jsonnet/blob/master/autoload/jsonnet.vim#L83

It would be great to improve these.

sparkprime:
Fix #278

### DIFF
--- a/core/formatter.cpp
+++ b/core/formatter.cpp
@@ -1081,6 +1081,23 @@ class FixIndentation {
         }
     }
 
+    /** Calculate the indentation of sub-expressions.
+     *
+     * If the first sub-expression is on the same line as the current node, then subsequent
+     * ones will be lined up and further indentations in their subexpresssions will be based from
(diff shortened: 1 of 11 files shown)


ahakanbaba:
Did this change go to the 0.9.1 ? 
If not could you create a new release with this ? 

Thanks

sparkprime:
It's done.

ahakanbaba:
Thanks.

Issue #300, opened 2017-02-11
Support Visual C++/Visual Studio toolchain


spencels:
Jsonnet and libjsonnet/libjsonnet++ should be buildable with the Visual C/C++ (VC/VC++) compiler. The project should include build files for a tool that integrates well with Visual Studio (VS) such as MSBuild, CMake, or NMake.

## Rationale

Microsoft's Visual C/C++ compiler is widely used on Windows. It's bundled with and integrated with Visual Studio, probably the most widely used IDE for C/C++ development on Windows. A small amount of development would give a wide community of developers easy access to Jsonnet.

Jsonnet currently supports Windows builds using Cygwin and MinGW, using make as a build system. These don't integrate well with VS. Cygwin .so libraries can't be used by other Windows development tools. VS can use DLLs and possibly even object files from MinGW's gcc, but not g++ as name mangling works differently. VS also doesn't integrate with the GNU toolchain: no make or gcc support (although there may be plugins for this). Developers wishing to use Jsonnet in a VC++ project must either forgo advanced IDE features like code search and debugging, or manually create their own project file.

VS is also the de facto IDE for .NET development. Supporting VS/VC++ will make writing and maintaining C#/.NET bindings for Jsonnet easier in the future.

thefoojoo:
I'm looking into this in my spare time, please assign to me if no one else is working on it.

sparkprime:
In principle, the code is all standard.  In practice, it may have bugs that only show up on VC.  I'm happy to accept PRs to fix these bugs :)  In 2014, I tried compiling Jsonnet with Visual Studio Express 12, but it crashed the compiler.  It is definitely worth trying it with a more modern MSVC.  There is this fork adding VC support which is quite old, but might provide some useful insight https://github.com/tyler-elric/jsonnet

Having yet another build system would be kind of a pain.  However I can't think of a better solution.  Is there some way of running regression tests on travis against VC?  The typical problem is that someone renames a header or adds a file or whatever, and only n-1 of the build scripts get updated, breaking one of the platforms.  Currently we have BUILD, Makefile, setup.py (for pypi).

thefoojoo:
I was able to compile Jsonnet and its tests without modification using VS2015. The question remains of what build system to use.

One build system option is to replace Make with [CMake](https://cmake.org/). It's a meta build system that generates targets for other build systems: Make, Visual Studio, Eclipse, etc. It also has features and plugins for other stuff like cloning external repos, copying files, and running external commands.

[...]

PR #303, opened 2017-02-27
CMake build and test.

spencels:
Add support for building and testing with CMake. This is necessary for future MSVC support (#300).

Not supported yet:
* Windows
* Emscripten
* Python
* Compilers other than GCC and Clang.


core/CMakeLists.txt
```diff
@@ -0,0 +1,67 @@
+# libjsonnet
+
+# Remember to update Bazel and Makefile builds when updating this list!
```
sparkprime: and MANIFEST.in and setup.py :(

sparkprime:
This looks ok, except it seems your build test is failing on mac?

I'd like to try it out myself, but haven't got time right now.

In particular, is there enough feature parity to remove the handwritten makefiles...

sparkprime:
Oh there is emscripten, as you said.

thefoojoo:
CMake could easily take over everything the handwritten makefile does now. The Python build be a CMake target, where CMake [generates setup.py/MANIFEST.in via a template](https://bloerg.net/2012/11/10/cmake-and-distutils.html) using the LIB_SOURCES variable in `core/CMakeLists.txt`. I believe there are CMake libraries out there for finding Python installations. I'm pretty sure CMake could also do the Emscripten build as well, but I'm less certain on how that would work. 

I'd prefer to implement Python and Emscripten builds in a follow-up PR to keep the size manageable, but I could build it into this one if you'd like.

I think Travis is failing on Linux. I developed this on a Mac so at least the Mac build is green :). I have access to all three systems so I will troubleshoot it when I get a chance.

thefoojoo:
I fixed the issue with Linux, Travis should work now (I'm not sure why it hasn't re-run). I also added a cross-platform std.jsonnet.h generator.

I noticed some functionality in run_tests.sh. I think I can implement that in CMake. Shell scripts won't run on Windows so I'll want to port that eventually.

sparkprime:
There is one script that makes sure jsonnet outputs "true" (or matches the golden file if provided) and another script that makes sure jsonnet fmt does not change the jsonnet file (or matches the golden file if provided).

sparkprime:
Hey, should I have merged this?

spencels:
Yes, I think this was ready to go out. I'll resolve merge conflicts and give it a quick test in the next week and let you know when it's ready.

sparkprime:
Awesome, thanks for coming back to it after all this time.

spencels:
The code looks good overall, but tests are failing after the pulling in 8 months-worth of changes. I'll ping this thread when I have it working again.

spencels:
The error was on my end. This is ready to go.

sparkprime:
Thanks!

PR #314, opened 2017-06-01
Add `prune` to stdlib

### LINKED ISSUE #313: Feature proposal: std.prune
For ksonnet I created a `prune` method that removes all empty object, arrays and nulls from a parent structure.  This can clean up "cruft" that is left over by initialized fields for mixins.

https://github.com/ksonnet/ksonnet-lib/blob/master/ksonnet.beta.1/util.libsonnet#L3

I'd be happy to submit a PR to add this to the std library if there is any desire.

jbeda:
Fixes #313


sparkprime:
Doesn't make depend add enough dependencies?

_Sent from my Motorola Nexus 6 using [FastHub](https://play.google.com/store/apps/details?id=com.fastaccess.github)_

jbeda:
I didn't see that!  But `makedepend` wasn't installed on my mac and I get a lot of warnings when I run it.  I'll just drop the makefile change for now from the PR here and tackle it separately.

PR #315, opened 2017-06-01
Fix `make depend` to look at right paths

jbeda:


jbeda:
This still throws a bunch of errors on the mac as makedepend assumes the location of the system header files.  Getting the include search paths out of gcc seems to be surprisingly hard.  There is nothing as easy as `go env` for gcc/g++.

sparkprime:
Thanks!

Issue #311, opened 2017-05-26
check if a field exists in 'super'


ant31:
Hi, 

I'm trying to do the following
```
local addMetadata(key, value) =
 {metadata+: {[key]: value};

{} + addMetadata("app", "name")
```
It fails because super doesn't have the 'metadata' field. I would like to know if there is a way to test that condition ?
eg:
```
if std.objectHas(super, 'metadata')
```

sparkprime:
#305 asks for the same thing.

My preferred approach is to put metadata: { } in the super object, but if that is not possible, we might have to revive the standalone super, which I removed because the implementation was very complex.

jbeda:
As we build out ksonnet, we want to be able to "mix in" things that the super object may not know about apriori.

An alternative would be to have something like `{} + initMetadata() + addMetadata(...)` but that is pretty verbose.

ant31:
Is this limited to a new notation or function would reduce complexity ? 

eg:
metadata|: {} 

or
std.superHas('metadata')

sparkprime:
So in general +: can be used to extend an array, string, or even to add to a number.  So a |: syntax sugar would have to default the super field to the correct "zero" value for the RHS of the operator.

What I used to do is have a template for my "base" objects that came from the library.  So you do:

`ksonnet.Base { } + ksonnet.mixin1 + ksonnet.mixin2` instead of `{ } + ksonnet.mixin1 + ksonnet.mixin2`

That allows you to add stuff to Base as you extend the library.  Can you do that?

jbeda:
@sparkprime this is much of what we are doing now.  The end result is that there is a lot of ugly cruft left over in the final rendered output (hence [prune](https://github.com/google/jsonnet/issues/313)).  As things get deeply nested this gets pretty big.  We also may find ourselves in a situation where we have:
```
lib1.Base + lib2.mixin1 + lib2.mixin2
```

`lib2` may be adding a field that `lib1` doesn't know about.  We'd then have to do something like:
```
lib1.Base + lib2.Base + lib2.mixin1 + lib2.mixin2
```

But that is pretty order dependent and a bit dangerous.

This triggers a thought: Another reason to allow detection on super would be to assert the thing you are clobbering is empty already.  That would prevent users from putting a base on top of something that already has data.

hausdorff:
[...]

PR #318, opened 2017-06-03
Permissive extends

sparkprime:
WIP on #311 

@hausdorff

### DIFF
--- a/core/desugarer.cpp
+++ b/core/desugarer.cpp
@@ -91,6 +91,9 @@ class Desugarer {
         return alloc->make<T>(std::forward<Args>(args)...);
     }
 
+    AST *clone(AST *ast)
+    { return clone_ast(*alloc, ast); }
+
     const Identifier *id(const UString &s)
     { return alloc->makeIdentifier(s); }
 
@@ -289,6 +292,10 @@ class Desugarer {
             }
         }
 
+        /** Replaces all occurrences of self, super[f] and e in super with variables.
+         *
+         * Returns all variables and original expressions via super_vars.
+         */
         class SubstituteSelfSuper : public CompilerPass {
             Desugarer *desugarer;
             SuperVars &superVars;
@@ -310,7 +317,7 @@ class Desugarer {
                     expr = alloc.make<Var>(expr->location, expr->openFodder, newSelf);
                 } else if (auto *super_index = dynamic_cast<SuperIndex*>(expr)) {
                     UStringStream ss;
-                    ss << U"$outer_super" << (counter++);
+                    ss << U"$outer_super_index" << (counter++);
                     const Identifier *super_var = desugarer->id(ss.str());
                     AST *index = super_index->index;
                     // Desugaring of expr should already have occurred.
@@ -318,6 +325,13 @@ class Desugarer {
                     // Re-use super_index since we're replacing it here.
                     superVars.emplace_back(super_var, super_index);
                     expr = alloc.make<Var>(expr->location, expr->openFodder, super_var);
+                } else if (auto *in_super = dynamic_cast<InSuper*>(expr)) {
+                    UStringStream ss;
+                    ss << U"$outer_in_super" << (counter++);
+                    const Identifier *in_super_var = desugarer->id(ss.str());
+                    // Re-use in_super since we're replacing it here.
+                    superVars.emplace_back(in_super_var, in_super);
+                    expr = alloc.make<Var>(expr->location, expr->openFodder, in_super_var);
                 }
                 CompilerPass::visitExpr(expr);
             }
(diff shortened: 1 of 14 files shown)


hausdorff:
Update, stuck in airport with broken wifi, and my flight just got cancelled. So review might be tomorrow.

core/desugarer.cpp
```diff
@@ -310,14 +317,21 @@ class Desugarer {
                     expr = alloc.make<Var>(expr->location, expr->openFodder, newSelf);
                 } else if (auto *super_index = dynamic_cast<SuperIndex*>(expr)) {
                     StringStream ss;
-                    ss << "$outer_super" << (counter++);
+                    ss << "$outer_super_index" << (counter++);
```
sparkprime: Yeah String and StringStream are both char32_t.  We should probably rename them to UStringStream and UString or something.

All internal string processing is in that format, however the I/O, i.e. lexing and manifesting are in UTF-8.  So there we convert between std::string and String with explicit encode/decode functions.  Also, the Identifier object used in parameters and variables (as in the case you called out here) is also stored in UTF-8 because it is limited to [a-zA-Z_][_a-zA-Z0-9]*.  They're also interned.  Both are arbitrary decisions.

Finally, the field names as stored in the HeapObject are Identifiers (even though they can be arbitrary strings).  This technically can lead to a high-level memory leak, because being interned, they are never free'd and the set of field names is not statically bound.  This doesn't cause a probably in practice, but we probably want to fix it in the Go version.

core/desugarer.cpp
```diff
@@ -331,14 +345,38 @@ class Desugarer {
...
-            AST *index = field.expr1;
             // Clone it so that we maintain the AST as a tree.
-            ClonePass(*alloc).expr(index);
+            AST *index = clone(field.expr1);
             // This will remove self/super.
             SubstituteSelfSuper(this, super_vars, counter).expr(index);
-            AST *super_f = make<SuperIndex>(field.expr1->location, EF, EF, index, EF, nullptr);
-            field.expr2 = make<Binary>(ast->location, EF, super_f, EF, BOP_PLUS, field.expr2);
-            field.superSugar = false;
+            if (true) {
```
sparkprime: Yeah this was a temporary experiment.  For `e+: e'`, The true branch is

```
[e]: (if e in super then super[e] else std.zero(std.type(e')) + e'
```

The false branch is

```
[e]: if e in super then super[e] + e' else e'
```

This relates to our semantic discussion:  The two branches are equivalent in a world where `e in super` never fails.  So going forward I'll use the false branch because it's simpler.  The difference is that in the true branch, + is involved in both the two conditional cases.  For mixins `Inner = { g+: { x: 3 } }` and `Outer = { f+: Inner }` in the desugaring of Outer we have `e' = Inner`.

In the false branch this would mean Outer is desugared to `After = { f: if "f" in super then super.f + Inner else Inner }`.  The result of that desugaring is that `{ } + After` would be an error because `Inner` cannot stand by itself (the "g" in super throws an error).

In the true branch the result of `{ } + After` not Inner standing alone but `std.zero(std.type(Inner)) + Inner`, which is `{ } + Inner` which means by induction it will work for all nestings.

However all of this is academic now since we agreed on the permissive super :)

Regarding your description of your understanding, both versions work with strings.  `{} + ""` will do an implicit string conversion of the `{}` and indeed create a "{}" string rather than a HeapExtendedObject.  However `{} + ""` only occurs in a case like `{ f: {} } + {f+: ""}` which for both cases should yield `{ f: "{}" }`, which is correct.  In vm.cpp `countLeaves(self)` is only ever used when self has been taken from the stack, so it is always some sort of HeapObject  (A HeapString is not a HeapObject, as the base type of all garbage collected things is HeapEntity).  The else branch in countLeaves is just looking for other HeapObjects, i.e. HeapLeafObject (they have a common supertype).

Thanks for taking the time to figure out what you did, we really need more people to be confident in this code.

core/pass.cpp
```diff
@@ -424,3 +440,10 @@ void ClonePass::expr(AST *&ast_)
 
     CompilerPass::expr(ast_);
 }
+
+AST *clone_ast(Allocator &alloc, AST *ast)
```
sparkprime: Just so it is easier to call it.  I realized we would probably be calling it quite a lot, and don't want to discourage cloning the subtree, because if the tree becomes a dag it leads to subtle bugs, whereas unnecessarily cloning is pretty harmless (slightly more static memory use).

core/vm.cpp
```diff
@@ -1839,16 +1849,49 @@ class Interpreter {
...
+                            if (lhs.t != Value::STRING) {
+                                throw makeError(ast.location,
+                                                "The left hand side of the 'in' operator should be "
+                                                "a string,  got " + type_str(lhs));
+                            }
+                            auto *field = static_cast<HeapString*>(lhs.v.h);
+                            switch (rhs.t) {
+                                case Value::ARRAY: {
+                                    auto *arr = static_cast<HeapArray*>(rhs.v.h);
+                                    // TODO(dcunnin): Implement this.
```
sparkprime: Yeah I remember you making that argument before.  I wouldn't consider having `x in [...]` except that we already have it in the array and object comprehensions (as they're the same syntax as Python).  In that case it's obviously O(n) though.

Another complication is we're representing sets as arrays (sets are a library that makes sorted arrays).  In that case `e in my_set` would be O(n) instead of O(log n) which is even more confusing.

That could be fixed by having sets as a first class value, which can be done backwards-compatible as long as `std.type(my_set)` is "array".  However that may be even more confusing :)

I think there's enough doubt to keep `e in arr` out for the moment at least.

core/vm.cpp
```diff
@@ -1839,16 +1849,49 @@ class Interpreter {
...
+                                case Value::OBJECT: {
+                                    auto *obj = static_cast<HeapObject*>(rhs.v.h);
+                                    auto *fid = alloc->makeIdentifier(field->value);
+                                    unsigned unused_found_at = 0;
+                                    bool in = findObject(fid, obj, 0, unused_found_at);
+                                    scratch = makeBoolean(in);
+                                } break;
+  
+                                default:
+                                throw makeError(ast.location,
```
sparkprime: I checked Python, it's annoying in that [1, 2, 3] is not in [0, 1, 2, 3, 4] but "123" is in "01234".  So let's punt string as well.

core/desugarer.cpp
sparkprime: Oh looks like I lied:  Identifiers are UTF32 which is why it uses StringStream in this case.  However the literals should also be UTF32 and I'm not sure why the compiler didn't complain here.  I guess because << is more permissive than std::base_string's + operator.  I will fix the literals.

Issue #210, opened 2016-06-24
jsonnet fmt should be able to take multiple file names


mikedanese:
To play nicely with bash glob. 

```
$ jsonnet -i **/*.jsonnet
```

sparkprime:
Fixed by #324

PR #324, opened 2017-07-13
Allow reformatting of multiple files

sbarzowski:
When using -i or --test.

See #210.

### DIFF
--- a/cmd/jsonnet.cpp
+++ b/cmd/jsonnet.cpp
@@ -72,9 +72,15 @@ void version(std::ostream &o)
 void usage(std::ostream &o)
 {
     version(o);
-    o << "Typical Usage:\n";
-    o << "jsonnet [<cmd>] {<option>} <filename>\n";
-    o << "Where <cmd> is one of {eval, fmt} and defaults to eval.\n";
+    o << "\n";
+    o << "General commandline:\n";
+    o << "jsonnet [<cmd>] {<option>} { <filename> }\n";
+    o << "Note: <cmd> defaults to \"eval\"\n";
+    o << "\n";
+    o << "The eval command:\n";
+    o << "jsonnet eval {<option>} <filename>\n";
+    o << "Note: Only one filename is supported\n";
+    o << "\n";
     o << "Available eval options:\n";
     o << "  -h / --help             This message\n";
     o << "  -e / --exec             Treat filename as code\n";
@@ -95,7 +101,6 @@ void usage(std::ostream &o)
     o << "Provide a value as Jsonnet code:\n";
     o << "  --ext-code <var>[=<code>]    If <code> is omitted, get from environment var <var>\n";
     o << "  --ext-code-file <var>=<file> Read the code from the file\n";
-    o << "\n";
     o << "Available options for specifying values of 'top-level arguments':\n";
     o << "Provide the value as a string:\n";
     o << "  -A / --tla-str <var>[=<val>]     If <val> is omitted, get from environment var <var>\n";
@@ -104,14 +109,16 @@ void usage(std::ostream &o)
     o << "  --tla-code <var>[=<code>]    If <code> is omitted, get from environment var <var>\n";
     o << "  --tla-code-file <var>=<file> Read the code from the file\n";
     o << "\n";
+    o << "The fmt command:\n";
+    o << "jsonnet fmt {<option>} { <filename> }\n";
+    o << "Note: Some options do not support multiple filenames\n";
+    o << "\n";
     o << "Available fmt options:\n";
-    o << "jsonnet fmt {<option>} <filename>\n";
-    o << "and <option> can be:\n";
     o << "  -h / --help             This message\n";
     o << "  -e / --exec             Treat filename as code\n";
     o << "  -o / --output-file <file> Write to the output file rather than stdout\n";
-    o << "  -i / --in-place         Update the Jsonnet file in place.  Same as -o <filename>\n";
-    o << "  --test                  Exit with failure if reformatting changed the file.\n";
+    o << "  -i / --in-place         Update the Jsonnet file(s) in place.\n";
+    o << "  --test                  Exit with failure if reformatting changed the file(s).\n";
     o << "  -n / --indent <n>       Number of spaces to indent by (default 0, means no change)\n";
     o << "  --max-blank-lines <n>   Max vertical spacing, 0 means no change (default 2)\n";
     o << "  --string-style <d|s|l>  Enforce double, single quotes or 'leave' (the default)\n";
@@ -152,7 +159,7 @@ enum Command {
 /** Class for representing configuration read from command line flags.  */
 struct JsonnetConfig {
     Command cmd;
-    std::string inputFile;
+    std::vector<std::string> inputFiles;
     std::string outputFile;
     bool filenameIsCode;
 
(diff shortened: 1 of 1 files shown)


sbarzowski:
My first meaningful change, please roast. Ping @lwander, @sparkprime.

Note: There are no tests at all yet for -i or --test options. I intend to add them in another commit.

cmd/jsonnet.cpp
```diff
@@ -618,27 +628,29 @@ int main(int argc, const char **argv)
...
-            jsonnet_destroy(vm);
-            return EXIT_FAILURE;
-        }
-
         // Evaluate input Jsonnet and handle any errors from Jsonnet VM.
         int error;
         char *output;
         switch (config.cmd) {
             case EVAL: {
+                assert(config.inputFiles.size() == 1);
```
lwander: I could be reading this wrong, but if `inputFiles` is empty the above check you guard with that "ERROR: Already specified" message won't catch a malformed userinput of 0 files.

cmd/jsonnet.cpp
```diff
@@ -673,38 +685,76 @@ int main(int argc, const char **argv)
...
-                        jsonnet_destroy(vm);
-                        return EXIT_FAILURE;
+
+                if (config.fmtInPlace || config.fmtTest) {
+                    assert(config.inputFiles.size() >= 1);
+                    for (std::string inputFile: config.inputFiles) {
+                        if (config.fmtInPlace) {
+                            output_file = inputFile;
+
+                            if (inputFile == "-") {
```
lwander: I realize you copied this restriction from the existing code, but why does it exist @sparkprime? Seems useful to pipe files through the formatter rather than modify them in place.

cmd/jsonnet.cpp
```diff
@@ -455,44 +456,53 @@ static bool process_args(int argc,
...
+        perror(msg.c_str());
         return false;
     }
-    config->inputFile = filename;
     return true;
 }
 
 /** Reads Jsonnet code from the input file or stdin into the input buffer. */
-static bool read_input(JsonnetConfig* config, std::string* input) {
+static bool read_input(JsonnetConfig* config, int index, std::string* input) {
```
lwander: Why did you introduce `index` as a parameter? I only see this function called [here](https://github.com/google/jsonnet/pull/324/files#diff-cbfdcc436b29a37aa318778257ca6871R640) and [here](https://github.com/google/jsonnet/pull/324/files#diff-cbfdcc436b29a37aa318778257ca6871R743), each time with `0` as `index`.

cmd/jsonnet.cpp
```diff
@@ -455,44 +456,53 @@ static bool process_args(int argc,
...
-        *input = config->inputFile;
-        config->inputFile = "<cmdline>";
+        *input = config->inputFiles[index];
+        config->inputFiles[index] = "<cmdline>"; // side effect!
     } else {
-        // Input file "-" tell Jsonnet to read stdin.
-        if (config->inputFile == "-") {
-            config->inputFile = "<stdin>";
+        // Input file "-" tells Jsonnet to read stdin.
+        if (config->inputFiles[index] == "-") {
```
lwander: Looking below I guess not - this block of code reads like it could, but I think once we figure out what to do with the constant `index` parameter this won't be a problem.

cmd/jsonnet.cpp
```diff
@@ -673,38 +685,76 @@ int main(int argc, const char **argv)
...
+
+                        std::string input;
+                        if (!read_input_contents(inputFile, &input)) {
+                            jsonnet_destroy(vm);
+                            return EXIT_FAILURE;
+                        }
+
+                        output = jsonnet_fmt_snippet(vm, inputFile.c_str(), input.c_str(), &error);
+
+                        if (error) {
```
lwander: It'd be nice if jsonnet didn't modify files until it was sure it could correctly format all of them. WDYT @sparkprime?

cmd/jsonnet.cpp
```diff
@@ -673,38 +685,76 @@ int main(int argc, const char **argv)
...
-
-                if (config.fmtTest) {
-                    // Check the output matches the input.
-                    bool ok = output == input;
-                    jsonnet_realloc(vm, output, 0);
-                    jsonnet_destroy(vm);
-                    return ok ? EXIT_SUCCESS : 2;
+                    if (error) {
+                        std::cerr << output;
+                        std::cerr.flush();
```
lwander: STDERR is unbuffered, I don't think this is needed: https://linux.die.net/man/3/stderr

cmd/jsonnet.cpp
```diff
@@ -455,44 +456,53 @@ static bool process_args(int argc,
...
+    }
+    config->inputFiles = remaining_args;
+    return true;
+}
+
+static bool read_input_contents(std::string filename, std::string *input) {
+    std::ifstream f;
+    f.open(filename);
+    if (!f.good()) {
+        std::string msg = "Opening input file: " + filename;
```
lwander: In other places the convention seems to be to write errors with `std:cerr`, any reason to use `perror` instead here?

sbarzowski: Well, it was like that before, I have no idea. Probably b/c that was shorter than using strerror and then cerr. @sparkprime?

cmd/jsonnet.cpp
sbarzowski: There is a check for that just above the one for "ERROR: Already specified".  I didn't need to touch it.

cmd/jsonnet.cpp
sbarzowski: > Seems useful to pipe files through the formatter rather than modify them in place.
Of course. It's only preventing specifying the input as "-" when changing the files in place (see the `if` above).
It's still possible to do `jsonnet fmt - < somefile > output_file`, it only prevents `jsonnet fmt --in-place -` which doesn't make any sense.

cmd/jsonnet.cpp
sbarzowski: The idea was that this function shouldn't care which of the many specified files it reads, but yes it ends up being used only when there is exactly one file.

cmd/jsonnet.cpp
```diff
@@ -455,44 +456,53 @@ static bool process_args(int argc,
...
 }
 
 /** Reads Jsonnet code from the input file or stdin into the input buffer. */
-static bool read_input(JsonnetConfig* config, std::string* input) {
+static bool read_input(JsonnetConfig* config, int index, std::string* input) {
     if (config->filenameIsCode) {
-        *input = config->inputFile;
-        config->inputFile = "<cmdline>";
+        *input = config->inputFiles[index];
+        config->inputFiles[index] = "<cmdline>"; // side effect!
```
sbarzowski: It's in the function named read_input, so I would expect getting and returning the contents to be the purpose of this function. But in addition to that, in special cases it changes the filename to sth else.

cmd/jsonnet.cpp
sbarzowski: Yes and no. Anyway you made me realize that there is a subtle regression - it's impossible now to use --test with stdin (it will treat it as a filename). I'll fix it.

cmd/jsonnet.cpp
sbarzowski: I see two ways to approach this:
a) Process everything as we do know, but keep the results in memory and write all of them at once afterwards.
b) Just do the initial validation on all files (check that they exist, that stdin is not used with -i, maybe that they can be opened in case permissions are wrong).

Of course, even with a), it's still possible that only some of them will be updated if something bad happens (file removed, permissions changed, process killed).

sbarzowski:
Thanks, I'm going to fix the issues in the morning.

cmd/jsonnet.cpp
lwander: It might be neater to take a pointer to the input file you're going to read then, unless you think there is an imminent use case for taking `index` that's not shown yet

cmd/jsonnet.cpp
lwander: Thanks, could you add this to the comment description?

cmd/jsonnet.cpp
lwander: Good catch

cmd/jsonnet.cpp
lwander: Ah I see, thanks

cmd/jsonnet.cpp
lwander: Yeah this is a little tricky, and might just be overkill. Curious to see what @sparkprime thinks

cmd/jsonnet.cpp
```diff
@@ -455,49 +456,73 @@ static bool process_args(int argc,
...
+            return false;
+        }
     }
-    config->inputFile = filename;
+    config->inputFiles = remaining_args;
     return true;
 }
 
+
 /** Reads Jsonnet code from the input file or stdin into the input buffer. */
```
sparkprime: Can you update the comment - the code does not care what is in the file.

cmd/jsonnet.cpp
sparkprime: I think it's fine to write out one at a time.

cmd/jsonnet.cpp
```diff
@@ -673,38 +699,75 @@ int main(int argc, const char **argv)
...
+                            jsonnet_destroy(vm);
+                            return EXIT_FAILURE;
+                        }
+
+                        if (config.fmtTest) {
+                            // Check the output matches the input.
+                            bool ok = output == input;
+                            jsonnet_realloc(vm, output, 0);
+                            if (!ok) {
+                                return 2;
```
sparkprime: Don't we need to destroy the vm here?

cmd/jsonnet.cpp
```diff
@@ -110,8 +110,8 @@ void usage(std::ostream &o)
     o << "  -h / --help             This message\n";
```
sparkprime: Can you change the help message to better promote the ability to use multiple filenames (I also tweaked some of the text here):

```
General commandline:
jsonnet [<cmd>] {<option>} { <filename> }
Note: <cmd> defaults to "eval".
```

```
The eval command:
jsonnet eval {<option>} <filename>
Note: Only one filename is supported
```

```
The fmt command:
jsonnet fmt {<option>} { <filename> }
Note: Some options do not support multiple filenames.
```

And then remove the line `Note: Multiple filenames can be provided at once when using -i or --test options.`

PR #325, opened 2017-07-19
Add an option for import sorting in fmt mode

sbarzowski:


sbarzowski:
Ping @lwander, @sparkprime. 
There are some minor unrelated changes, namely the addition of ASTTypeToString which is very helpful when debugging AST and some trailing whitespace that my editor kindly removed. Of course I can move these things to a separate commit if you think it's worth it.

core/formatter.cpp
```diff
@@ -594,6 +595,26 @@ class Unparser {
  * The rest of this file contains transformations on the ASTs before unparsing. *
  ********************************************************************************/
 
+/** As a.push_back(elem) but preserves constraints.
```
lwander: Can you extend the comment to explain "constraints"?

core/formatter.cpp
```diff
@@ -1724,16 +1733,233 @@ class FixIndentation {
...
+        }
+    }
+};
 
 std::string jsonnet_fmt(AST *ast, Fodder &final_fodder, const FmtOpts &opts)
 {
     Allocator alloc;
 
     // Passes to enforce style on the AST.
+    if (opts.sortImports)
```
lwander: Little nit, but I think here it's generally recommended to include braces for one line if-statements (I could be wrong)

core/formatter.cpp
```diff
@@ -1724,16 +1733,233 @@ class FixIndentation {
...
+    bool duplicatedVariables(const ImportElems &elems) {
+        std::vector<UString> idents;
+        for (const auto &elem: elems) {
+            idents.push_back(elem.bind->var->name);
+        }
+        return std::unique(idents.begin(), idents.end()) != idents.end();
+    }
+
+    /// Check if the import group ends after this local
+    bool groupBoundaryAfter(Local *local) {
```
lwander: This doesn't modify the input `local` contents, right? The current name kind-of suggests that it does - maybe `groupBoundaryStartsAfter` sounds more like you're asking a question rather than asking it to group a boundary.

core/formatter.cpp
```diff
@@ -1724,16 +1733,233 @@ class FixIndentation {
...
+        }
+    }
+
+    AST *toplevelImport(Local *local, ImportElems &imports, const Fodder &groupOpenFodder) {
+        assert(isGoodLocal(local));
+
+        Fodder commentAfter, beforeNextFodder;
+        std::tie(commentAfter, beforeNextFodder) = splitOpenFodder(local->body);
+
+        ensureCleanNewline(commentAfter);
```
lwander: What if you have multiple comments? e.g 

```
local foo = "b.jsonnet"; /* comment1 */ /* comment2 */
```

core/formatter.cpp
```diff
@@ -1724,16 +1733,233 @@ class FixIndentation {
...
+            }
+
+            return buildGroupAST(imports, bodyAfterGroup, groupOpenFodder);
+        } else {
+            assert(beforeNextFodder.empty());
+            return toplevelImport(dynamic_cast<Local*>(local->body), imports, groupOpenFodder);
+        }
+    }
+
+    void file(AST *&body)
```
lwander: Looks like other class methods have the brace on the same line as the method declaration

core/formatter.cpp
```diff
@@ -1724,16 +1733,233 @@ class FixIndentation {
...
+            }
+        }
+        return {afterPrev, beforeThis};
+    }
+
+    void sortGroup(ImportElems &imports) {
+        // We don't want to change behaviour in such case:
+        // local foo = "b.jsonnet";
+        // local foo = "a.jsonnet";
+        // So we don't change the order when there are shadowed variables.
```
lwander: Nice

sbarzowski:
This failure is not a failed test, but some configuration issue/external service failure:
```
Downloading https://get.rvm.io
Could not download rvm-installer, get some help at #rvm IRC channel at freenode servers.
```
BTW is it possible to run the failed check again manually?

core/formatter.cpp
sbarzowski: It's the same as in the function below, I am referring to it in "See concat_fodder". It's short, so maybe duplicating it won't be so bad. Or maybe there is a better place to put these constraints?

core/formatter.cpp
```diff
@@ -1724,16 +1733,233 @@ class FixIndentation {
...
+        for (auto &bind: local->binds) {
+            Import *import = dynamic_cast<Import*>(bind.body);
+            assert(import != nullptr);
+            result.emplace_back(sortingKey(import), after, &bind);
+            after = {FodderElement(FodderElement::Kind::LINE_END, 0, 0, {})};
+        }
+        return result;
+    }
 
+    AST *buildGroupAST(ImportElems &imports, AST *body, const Fodder &groupOpenFodder) {
```
sbarzowski: Well, existing code often goes over 80, so I didn't worry about it very much. @sparkprime?

lwander:
You can try `git commit --amend` to change your commit hash, and then `git push -f  origin <branch>` to kick of the build again when github sees the new commit.

lwander:
I think there are permissions that can be set within Travis-CI to allow us to manually restart the build, but I'm not sure who to contact for this repo to grant us those permissions.

core/formatter.cpp
sbarzowski: Maybe commentAfter is not the best name. Any suggestions? It contains all comments until the end of the line. In case a comment goes beyond the end of the line it's also a part of commentAfter. However if there are some blank lines afterwards, they go to beforeNextFodder.

Example: 
```
local foo = 'a.jsonnet'; /* blah blah*/ /* blah blah
blah blah

*/

// something else

{}
```
commentAfter =
```
/* blah blah*/ /* blah blah
blah blah

*/
```
beforeNextFodder ==
```

// something else

```

core/formatter.cpp
lwander: Ah if its in the surrounding code I'd leave it as is

core/formatter.cpp
sbarzowski: Well, it's not the boundary that starts :-). So maybe just `groupEndsAfter`?

core/formatter.cpp
lwander: Ah nice - I guess the name is a little tricky, maybe `adjacentCommentFodder`?

core/formatter.cpp
sbarzowski: :+1:, I like to always put braces. However similar code in this function doesn't use them. @sparkprime?

core/formatter.cpp
lwander: Sounds good to me

core/formatter.cpp
sbarzowski: Most (but not all) methods in this file have it on a new line, so I'll convert to that. Maybe I should try running clang-format over the whole code (not this change ofc) and maybe I can add it to automatic tests so that we have it always consistent.

core/lexer.h
```diff
@@ -56,7 +56,11 @@ static inline std::ostream &operator<<(std::ostream &o, const FodderElement &f)
 {
     switch (f.kind) {
         case FodderElement::LINE_END:
-        o << "END(" << f.blanks << ", " << f.indent << ", " << f.comment[0] << ")";
+        o << "END(" << f.blanks << ", " << f.indent;
+        if (!f.comment.empty()) {
+            o << ", " << f.comment[0];
+        };
```
sbarzowski: stray semicolon

core/formatter.cpp
lwander: Good idea

sbarzowski: Done, i.e. it's consistent with the rest of the file now.

core/formatter.cpp
sbarzowski: Ok, I got the answer. It's 100 columns.

core/ast.h
```diff
@@ -60,6 +60,101 @@ enum ASTType {
...
+        return "AST_SUPER_INDEX";
+        break;
+        case AST_UNARY:
+        return "AST_UNARY";
+        break;
+        case AST_VAR:
+        return "AST_VAR";
+        break;
+        default:
+        assert(false && "Unknown AST type");
```
sbarzowski: Well, "???" was there just to make the compiler happy. There's an `assert(false && "err msg")` just above it, doing pretty much the same thing as your proposition.

Next time I'll add a comment `// never reached` or sth like that to the `return` statement, but here the compiler is smart enough not to issue a warning about not returning if I just remove `return "???"` (probably b/c it can check that all cases are covered). 

Please let me know if you think that:
```
std::cerr << "Unknown AST type" << std::endl;
abort();
```
is better than:
```
assert(false && "Unknown AST type");
```

core/formatter.cpp
```diff
@@ -1724,16 +1734,245 @@ class FixIndentation {
...
+        return body;
+    }
 
+    bool duplicatedVariables(const ImportElems &elems)
+    {
+        std::vector<UString> idents;
+        for (const auto &elem: elems) {
+            idents.push_back(elem.bind->var->name);
+        }
+        return std::unique(idents.begin(), idents.end()) != idents.end();
```
sbarzowski: Argh, I forgot to sort it. :/ I believe `sort` + `unique` is faster, so I default to it, even though it's (as I have demonstrated) a little bug-prone. But yeah, I think set solution is more readable, I'll change to that.

core/ast.h
```diff
@@ -60,6 +60,101 @@ enum ASTType {
     AST_VAR
 };
 
+static inline std::string ASTTypeToString(ASTType type) {
+    switch(type) {
+        case AST_APPLY:
+        return "AST_APPLY";
+        break;
```
sbarzowski: done

core/ast.h
sparkprime: It is better because the asserts can be disabled, and most people expect that to not change the behavior of the code.

core/formatter.cpp
```diff
@@ -1724,16 +1734,245 @@ class FixIndentation {
...
+        // We don't want to change behaviour in such case:
+        // local foo = "b.jsonnet";
+        // local foo = "a.jsonnet";
+        // So we don't change the order when there are shadowed variables.
+        if (!duplicatedVariables(imports)) {
+            std::sort(imports.begin(), imports.end());
+        }
+    }
+
+    ImportElems extractImportElems(Local *local, Fodder after)
```
sbarzowski: Done, i.e. I have fixed it. I may add some more tests.

BTW I see some real value in it - in discoverability/UX. Otherwise the first-time user may assume that import sorting doesn't work or that it requires some additional setup and never change the style (how would they know that they need to use one-local-per-import?).

core/ast.h
sparkprime: Actually can't you just remove the default case entirely?

sbarzowski: done

core/formatter.cpp
```diff
@@ -1724,16 +1734,245 @@ class FixIndentation {
...
+    /// The imports are sorted by UTF-32 codepoints without case folding
+    UString sortingKey(Import *import)
+    {
+        return import->file->value;
+    }
+
+    /// Check if `local` expression is used for importing,
+    bool isGoodLocal(Local *local)
+    {
+        for (const auto &bind: local->binds) {
```
sbarzowski: done (+ test)

core/formatter.cpp
```diff
@@ -1724,16 +1734,245 @@ class FixIndentation {
...
+ * that are either the root of AST or a direct child (body) of a top-level
+ * import.
+ *
+ * Grouping of imports is preserved. Groups of imports are separated by blank
+ * lines or lines containing comments.
+ */
+class SortImports {
+    /// Internal representation of an import
+    struct ImportElem {
+        ImportElem(UString key, Fodder afterFodder, Local::Bind *bind)
```
sbarzowski: done

sbarzowski: done

core/formatter.cpp
```diff
@@ -1724,16 +1734,245 @@ class FixIndentation {
...
+            if (inSecondPart) {
+                fodder_push_back(beforeThis, fodderElem);
+            } else {
+                afterPrev.push_back(fodderElem);
+            }
+            if (fodderElem.kind != FodderElement::Kind::INTERSTITIAL) {
+                inSecondPart = true;
+                if (fodderElem.blanks > 0) {
+                    afterPrev.back().blanks = 0;
+                    assert(beforeThis.empty());
```
sbarzowski: good catch, fixed

core/formatter.cpp
```diff
@@ -1724,16 +1734,245 @@ class FixIndentation {
...
+            if (i == 0) {
+                fodder = groupOpenFodder;
+            } else {
+                fodder = imports[i - 1].afterFodder;
+            }
+            Local::Binds binds({*import.bind});
+            auto *local = alloc.make<Local>(
+                LocationRange(),
+                fodder,
+                binds,
```
sbarzowski: No, I can't. But a separate variable is unnecessary.

core/formatter.cpp
```diff
@@ -1724,16 +1734,245 @@ class FixIndentation {
...
+        // local foo = "a.jsonnet";
+        // So we don't change the order when there are shadowed variables.
+        if (!duplicatedVariables(imports)) {
+            std::sort(imports.begin(), imports.end());
+        }
+    }
+
+    ImportElems extractImportElems(Local *local, Fodder after)
+    {
+        ImportElems result;
```
sbarzowski: I have changed the function to just take the binds, instead of whole local, so now it should be obvious.

core/formatter.cpp
```diff
@@ -1724,16 +1734,245 @@ class FixIndentation {
...
+    {
+        if (auto *local = dynamic_cast<Local*>(expr)) {
+            return isGoodLocal(local) ? local : nullptr;
+        } else {
+            return nullptr;
+        }
+    }
+
+    /** Split openFodder into what logically belongs to the previous AST node
+     * what belongs to the current AST node.
```
sbarzowski: done (there are major changes to this comment)

core/formatter.cpp
```diff
@@ -1724,16 +1734,245 @@ class FixIndentation {
...
+     *
+     * Example:
+     * local x = import "xxx.jsonnet"; // Importing stuff here
```
sbarzowski: Done. I went with prev_token, next_token.

core/formatter.cpp
```diff
@@ -1724,16 +1734,245 @@ class FixIndentation {
...
+     *
+     * // blah blah
+     * {}
+     *
+     * In such case "// Importing  stuff here\n" part of the fodder belongs
+     * to the `local` and the rest to the {}.
+     */
+    std::pair<Fodder, Fodder> splitOpenFodder(AST *expr)
```
sbarzowski: Done.

core/formatter.cpp
```diff
@@ -1724,16 +1734,245 @@ class FixIndentation {
...
+        for (const auto &fodderElem: fodder) {
+            if (inSecondPart) {
+                fodder_push_back(beforeThis, fodderElem);
+            } else {
+                afterPrev.push_back(fodderElem);
+            }
+            if (fodderElem.kind != FodderElement::Kind::INTERSTITIAL) {
+                inSecondPart = true;
+                if (fodderElem.blanks > 0) {
+                    afterPrev.back().blanks = 0;
```
sbarzowski: done

core/formatter.cpp
sbarzowski: eh, there are still some weird cases:
```
local /* bbb */ b = import "b.jsonnet",  # B comment.
      /* aaa
         aaa
         aaa */ a = import "a.jsonnet";  # A comment.
```
or
```
local b = import "b.jsonnet",  /* BBB */ /* B comment.
      bbb
      bbb
      bbb */ a = import "a.jsonnet";  # A comment.
```

sbarzowski: Ok, that's due to paragraph handling. Now the clean line invariant should be preserved. The output is still a little weird, but the input was too.

sbarzowski:
I think I have addressed all the mentioned issues.

There are two things that I would consider TODO afterwards, but the change is already big so I think separate commits will be better (and easier to review) if we want to do these things:
a) Refactor Fodder into a real class that preserves the invariants. We already have fodder_* "methods", but they're not used consistently. This seems easy, but it will change a lot of code.
b) After some thought I no longer like treating a paragraph comment as a group separator. This leads to a strange situation when if my comment is a little bit too long for LINE_END I need to split import group. This feels uncomfortable. But I think it can wait.

core/lexer.h
```diff
@@ -113,13 +117,76 @@ static inline std::ostream &operator<<(std::ostream &o, const FodderElement &f)
 
 /** A sequence of fodder elements, typically between two tokens.
  *
- * A LINE_END is not allowed to follow a PARAGRAPH or a LINE_END.  This can be represented by
+ * A LINE_END is not allowed to follow a PARAGRAPH or a LINE_END. This can be represented by
  * replacing the indent of the prior fodder and increasing the number of blank lines if necessary.
  * If there was a comment, it can be represented by changing the LINE_END to a paragraph containing
  * the same single comment string.
+ *
+ * There must be a LINE_END or a PARAGRAPH before a PARAGRAPH.
```
sparkprime: Didn't we decide this wasn't necessary?

sbarzowski: I think that actually this is a good thing after all.

Some reasons starting paragraphs on a clean line is a good thing (I think you've mentioned most of them in our discussion):
* Having paragraphs starting mid-line is a weird input and boundary case for all the passes that touch fodders and newlines. It's more "hygienic" to eliminate these cases completely.
* We don't allow paragraphs to end mid-line (which is a very good thing, as it would lead to even weirder cases and it would prevent clean conversion to single-line comments). So symmetry here is nice.
* I have randomly asked some people and apparently no one can think of a reason why someone would want to have such thing in their code.

Some reasons why the current representation is reasonable (compared to e.g. treating the line before as part of the paragraph):
* How to otherwise represent the line comment that can go in the line before? Keeping it in a paragraph is very ugly (it effectively gets rid of the invariant + allows mixing comment styles within a paragraph). Some rule like "PARAGRAPH starts with an endline unless there is LINE_END before it" is better, but still a little confusing.
* Why not just another pass? Because having these newlines affects other passes, for example FixIndentation and SortImports . So we want to keep this as invariant at all times and not just when we're done. 
* Some existing code depends on it. There's a good chance some bugs would be introduced while changing it. I'm not really convinced that the alternatives I can think of now are actually better so keeping things as they are seems like a reasonable default. Especially since this PR is about something else.

Issue #143, opened 2016-03-26
Write .clang-format for C++ style


davidzchen:
Related to #142, add a .clang-format that codifies the C++ style for Jsonnet.

sparkprime:
I'd be OK with changing the style if .clang-format is not able to enforce the way it currently is, so we should do this first, and then document the resulting state.

Presumably, it'll be possible to do run the C++ reformatter over the whole code base and update it, if that is the case?

davidzchen:
Sounds good to me. We can script [clang-format](http://clang.llvm.org/docs/ClangFormat.html) to automatically reformat all the C++ files in the codebase using the style defined in the .clang-format file.

Once we have this done, we an also use [clang-tidy](http://clang.llvm.org/extra/clang-tidy/) to run lint checks, similar to what we do internally. :)

PR #326, opened 2017-07-20
Reformat code (and add make target to do so)

sbarzowski:


sbarzowski:
it's a massive change, even though I tried to match the settings with the existing style. There were no manual changes (it's a bad idea to mix automatic reformatting with other restructuring imo). 

Sometimes it goes from bad to worse, for example when there are plenty of arguments to a function. But I think it's still a net win, especially if we enforce it automatically later on.

@lwander, @sparkprime

Makefile
```diff
@@ -101,6 +101,9 @@ test: jsonnet libjsonnet.so libjsonnet_test_snippet libjsonnet_test_file
 	cd test_suite ; ./run_tests.sh
 	cd test_suite ; ./run_fmt_tests.sh
 
+reformat:
```
lwander: `all` should probably depend on this.

also I think you should declare this as [`.PHONY`](https://www.gnu.org/software/make/manual/html_node/Phony-Targets.html)

core/static_analysis.cpp
```diff
@@ -31,44 +31,45 @@ static void append(IdSet &r, const IdSet &s)
 /** Statically analyse the given ast.
  *
  * \param ast_ The AST.
- * \param in_object Whether or not ast_ is within the lexical scope of an object AST.
- * \param vars The variables defined within lexical scope of ast_.
- * \returns The free variables in ast_.
+ * \param in_object Whether or not ast_ is within the lexical scope of an object
+ * AST. \param vars The variables defined within lexical scope of ast_. \returns
```
lwander: This kind of doc reformatting seems wrong

.clang-format
```diff
@@ -0,0 +1,108 @@
+---
+Language:        Cpp
+# BasedOnStyle:  Google
```
lwander: Why commented out?

core/static_analysis.cpp
sbarzowski: I think I can fix it through CommentPragma option, thanks for noticing

Makefile
sbarzowski: I'm worried that reformatting on every build (== adding to `all`) may be annoying. 

:+1: for `.PHONY` I forgot about it. 

Also I guess I'll add a separate `test-formatting` target for use in TravisCI etc.

.clang-format
sbarzowski: The real reason is that it was autogenerated this way. But I guess I can either remove it completely or leave it active. I'll take another look at this config.

sparkprime:
https://github.com/google/jsonnet/issues/143

@davidzchen will be happy about this

.clang-format
```diff
@@ -0,0 +1,108 @@
...
+SpaceBeforeParens: ControlStatements
+SpaceInEmptyParentheses: false
+SpacesBeforeTrailingComments: 2
+SpacesInAngles:  false
+SpacesInContainerLiterals: true
+SpacesInCStyleCastParentheses: false
+SpacesInParentheses: false
+SpacesInSquareBrackets: false
+Standard:        Auto
+TabWidth:        8
```
sparkprime: Should this be 4?

sbarzowski: No, this is not the indentation, but how we interpret actual tab characters (we change them to this many spaces then). Google style uses 8 here, even though the Indentation is only 2 chars wide.

sbarzowski:
BTW if we want to enforce it automatically there are some workflow issues to discuss. Most importantly the problems with having different versions of clang-format on different machines which may disagree with the "official" version on TravisCI. The most proper solution I think is to pin down the version of clang-format somehow, locally for the project. I was using a workflow like this very successfully in Python and it was very easy with virtualenv. Some people do it for C++, but it's quite complex (https://engineering.mongodb.com/post/succeeding-with-clangformat-part-2-the-big-reformat) I suppose typically C++ projects use autoformatting in a more ad-hoc way.

sparkprime:
Did you try it with 100 line width yet?

sparkprime:
This doesn't seem too onerous:

The Wrapper Script

In any non-trivial use of clang-format, you will want a wrapper around it. Our script, clang_format.py:

Fetches clang-format binaries, from LLVM when it can, and from a MongoDB cache when it cannot
Validates clang-format is at the right version when binaries are already present
Skips files that we shouldn’t bother formatting, i.e. 3rd party library code, like boost
Offers modes for both validating and updating
Manages parallel invocations of clang-format to increase throughput

sbarzowski:
> Did you try it with 100 line width yet?
Yes, current version uses 100 as a target width.

Makefile
sparkprime: I don't think make all should change source code.  However it can run the test-formatting.

core/lexer.h
```diff
@@ -56,14 +56,14 @@ static inline std::ostream &operator<<(std::ostream &o, const FodderElement &f)
 {
     switch (f.kind) {
         case FodderElement::LINE_END:
-        o << "END(" << f.blanks << ", " << f.indent << ", " << f.comment[0] << ")";
-        break;
+            o << "END(" << f.blanks << ", " << f.indent << ", " << f.comment[0] << ")";
```
sparkprime: I have a moderate preference for keeping them on the same line, because they're essentially just indirect goto labels.  However if there's no supported option in clang for that then I can live with it.

core/lexer.h
```diff
@@ -144,66 +144,109 @@ struct Token {
...
-            case LOCAL: return "local";
-            case NULL_LIT: return "null";
-            case SELF: return "self";
-            case SUPER: return "super";
-            case TAILSTRICT: return "tailstrict";
-            case THEN: return "then";
-            case TRUE: return "true";
-
-            case END_OF_FILE: return "end of file";
+            case BRACE_L:
```
sparkprime: Any way to encourage it to put the return on the same line?  This is less readable than before.

core/lexer_test.cpp
```diff
@@ -19,20 +19,17 @@ limitations under the License.
 #include <list>
 #include "gtest/gtest.h"
 
-namespace {
-
-void testLex(const char* name,
-             const char* input,
-             const std::list<Token>& tokens,
+namespace
```
sparkprime: I've never seen anyone put the { after a namespace before

sparkprime: (on the line after)

core/lexer.h
sbarzowski: I checked that and I couldn't find anything to control that.

core/lexer.h
sbarzowski: Yes I experimented with that, I think there were fewer changes if I set it this way, but I might be wrong. Anyway, I'll change it.

core/libjsonnet.cpp
```diff
@@ -261,13 +264,17 @@ static char *default_import_callback(void *ctx, const char *dir, const char *fil
...
-#define CATCH(func) \
-    } catch (const std::bad_alloc &) {\
-        memory_panic(); \
-    } catch (const std::exception &e) {\
-        std::cerr << "Something went wrong during " func ", please report this: " \
-                  << e.what() << std::endl; \
-        abort(); \
+#define CATCH(func)                                                                           \
+    }                                                                                         \
+    catch (const std::bad_alloc &)                                                            \
```
sparkprime: I think try { } catch should have the same rules as if { } else { }

core/libjsonnet.cpp
```diff
@@ -277,16 +284,14 @@ const char *jsonnet_version(void)
 
 JsonnetVm *jsonnet_make(void)
 {
-    TRY
-    return new JsonnetVm();
+    TRY return new JsonnetVm();
```
sparkprime: uh-oh.

sparkprime: we could change this to
```
TRY {
    return new JsonnetVm();
} CATCH("jsonnet_make")
```

It'll probably behavior better then.

sparkprime: (move the { } out of the macro, that is)

core/libjsonnet.cpp
```diff
@@ -434,44 +440,40 @@ static char *jsonnet_fmt_snippet_aux(JsonnetVm *vm, const char *filename, const
...
-        if (!f.good()) {
-            std::stringstream ss;
-            ss << "Opening input file: " << filename << ": " << strerror(errno);
-            *error = true;
-            return from_string(vm, ss.str());
-        }
-        std::string input;
-        input.assign(std::istreambuf_iterator<char>(f),
-                     std::istreambuf_iterator<char>());
+    TRY std::ifstream f;
```
sparkprime: TRY { would fix this indentation too

core/libjsonnet.cpp
sbarzowski: It does. But here we have crazy macros.

core/libjsonnet.cpp
sbarzowski: I was able to tell clang-format that TRY ... CATCH is a block (as it is, no changes to the macro).

cmd/jsonnet.cpp
```diff
@@ -277,26 +275,26 @@ static bool process_args(int argc,
                     dir += '/';
                 }
                 jsonnet_jpath_add(vm, dir.c_str());
-            // TODO(dcunnin): Remove deprecated --var, --env and -E
-            } else if (arg == "-V" || arg == "--ext-str"
-                       || arg == "--var" || arg == "--env" || arg == "-E") {
+                // TODO(dcunnin): Remove deprecated --var, --env and -E
```
sparkprime: Maybe we should do this: #327

core/lexer.cpp
```diff
@@ -71,32 +70,29 @@ static void lex_ws(const char *&c, unsigned &new_lines, unsigned &indent, const
...
+                break;
 
             // This only works for \t at the beginning of lines, but we strip it everywhere else
             // anyway.  The only case where this will cause a problem is spaces followed by \t
             // at the beginning of a line.  However that is rare, ill-advised, and if re-indentation
             // is enabled it will be fixed later.
-            case '\t':
-            indent += 8;
-            break;
+            case '\t': indent += 8; break;
```
sparkprime: weird that it put it on the one line for this case but not the '  ' case above

core/lexer.cpp
```diff
@@ -208,137 +214,163 @@ std::string lex_number(const char *&c, const std::string &filename, const Locati
...
+                    case '0':
+                    case '1':
+                    case '2':
+                    case '3':
+                    case '4':
+                    case '5':
+                    case '6':
+                    case '7':
+                    case '8':
+                    case '9': state = AFTER_DIGIT; break;
```
sparkprime: Any way to get it to binpack the cases?

sparkprime:
Ok I read the whole thing.  Generally this is a big net improvement.  The majority of changes are actually good, the rest are things that don't make it better / worse but were not consistent across the whole codebase originally.  Probably I'd just do something random in each case.  The only thing that's still worth looking into is the bin packing of cases in that floating point number lexing rule.  However that's not a deal breaker by any means.  Thanks for setting this up!

core/lexer.cpp
sbarzowski: `clang-format` can sometimes produce weird results. Some quick experimentation shows that the reason why the `' '` case is not put in one line is the comment below it. But I have no idea why.

core/lexer.cpp
sbarzowski: I can't find any.

cmd/jsonnet.cpp
sbarzowski: Yes, that will help.

sbarzowski:
(rebased)

PR #329, opened 2017-07-25
Add idempotence tests

sbarzowski:
Just run formatter again on golden and see if something changes.
It shouldn't.

### DIFF
--- a/test_suite/run_fmt_idempotence_tests.sh
+++ b/test_suite/run_fmt_idempotence_tests.sh
@@ -1,0 +1,56 @@
+#!/bin/bash
+
+# Copyright 2015 Google Inc. All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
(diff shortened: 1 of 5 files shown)


sbarzowski:
This commit is a side-effect of my fights with weird paragraph fodder problems. Cool part is that it uses existing test data, so we have more comprehensive test-suite with little additional effort once we merge this. It's important to keep formatter idempotent, yet even very serious ones (like gofmt) have occasional problems with that. So IMO it's worth it to check.

@lwander @sparkprime

sparkprime:
:+1:

test_suite/run_fmt_idempotence_tests.sh
```diff
@@ -0,0 +1,56 @@
...
+    GOLDEN_OUTPUT="${ORIG_OUTPUT}"
+    GOLDEN_KIND="PLAIN"
+
+    if [ -r "$TEST.fmt.golden" ] ; then
+        GOLDEN_OUTPUT=$(cat "$TEST.fmt.golden")
+        FMT_FILE="$TEST.fmt.golden"
+    fi
+
+    if [ $(echo "$TEST" | cut -b 1-12) == "error.parse." ] ; then
+        continue  # No point testing these
```
lwander: Maybe it's worth testing that no regression in parsing/import formatting gets introduced that causes these to misbehave?

sbarzowski: We don't get to format `error.parse.*` at all (formatting happens after parsing), so there is no interaction with the formatter. And we check that the error is what we expect in run_tests.sh. So I think we're covered here.

Unless what you meant is running golden files and verifying that the output matches the output of original files. That could help with things like variable shadowing in imports:
```
local b = import "b.jsonnet";
local a = import "a.jsonnet";
local b = import "a.jsonnet";
```
But I guess that would be another script.

PR #333, opened 2017-07-27
Fix the bazel python build

benley:
For some reason `sysconfig.get_path("include")` incorrectly returns /usr/local/include/python2.7 on Debian (and probably Ubuntu) systems. Fortunately, `distutils.sysconfig.get_python_inc()` seems to do the right thing across platforms.


sparkprime:
Thanks!

Issue #299, opened 2017-02-03
Collection entry indentation inconsistency in fmt 


ahakanbaba:
Consider the following indentations auto generated by jsonnet fmt 

```
{
    collCorrect1: [{
        name: "Good",
    }],
    collCorrect2: [{
        name: "Bad",
    }, someOther(),
    ],
    collWeird: [{
                    name: "Bad",
                },
                someOther(),
    ],
}

```

It looks like the someOther() function being on a new line confuses the indentation calculation for the collWeird example. 

Would you consider this as a bug @sparkprime ?

ahakanbaba:
@huggsboson  FYI

sparkprime:
What do you expect it to do?  Personally I find collCorrect2 to look the weirdest because there is a } immediately followed by a ] at the same indentation level.

For collWeird, I say that if you choose not to start a new line, you're expecting everything to be lined up.  The ] is closing the initial [ so is not subject to that environment created within the [ ].

sparkprime:
Oh and yeah, in the case of someOther() being on the same line as the comma before it... in that case there is no element of the list with a linebreak preceding it, which is why it has chosen to not indent it (same as collCorrect1, but then there was only 1 element of the list so it is more obvious).

sparkprime:
It's kinda hard to codify rules like this...

huggsboson:
I would land on this:
```
{
    collCorrect2: [
        {
            name: "Bad",
        }, 
        someOther(),
    ]
```

I've largely started eschewing the [{ and }] on the same line as I keep running into edge cases like this and end up landing on.
```
[
    {
       ...
    },
    ...
]
```

Which takes up way more space but at least formats consistently.

ahakanbaba:
I should have been more clear in the behavior I see as surprising. 

I may be wrong but please bear with me. 
The collection has two entries separated with a comma. 

It looks to me that the second entry's location with respect to the comma affects the indentation of the first entry. That was surprising to me. Please let me know if your think that is expected behavior. 


Almost the same example in a more directed way and with line numbers : 

```
25     coll2: [{
26         name: "name",
27     }, someOther(),
28     ],
29     coll3: [{
30                 name: "Bad",
31             },
32             someOther(),
33     ],
34
```


> Personally I find collCorrect2 to look the weirdest because there is a } immediately followed by a ] at the same indentation level.

You may be right. I should not have used "Weird" in the key name. Maybe the coll2 should look like this ? 
25     coll2: [{
[...]

PR #334, opened 2017-07-28
Add simple newline insertion to formatter

sbarzowski:
This fixes "bracebug" (reported here: #299) and many
variants of it.

Of course it's not a complete solution to any newline problems.
In particular it never removes or rearranges newlines, it only
adds some. But it still should help a lot.

### DIFF
--- a/core/formatter.cpp
+++ b/core/formatter.cpp
@@ -1159,10 +1159,10 @@ class FixNewlines: public FmtPass {
     }
 
     // Example:
-    // very_long_function_on_a_long_line_so_arguments_dont_fit_on_the_same_line(
+    // foo(
     //     1, 2, 3)
     // Should be expanded to:
-    // very_long_function_on_a_long_line_so_arguments_dont_fit_on_the_same_line(
+    // foo(
     //     1, 2, 3
     // )
     bool shouldExpandNearParens(ArgParams &params, Fodder &fodder_r)
--- a/core/formatter.cpp
+++ b/core/formatter.cpp
@@ -963,6 +963,280 @@ class PrettyFieldNames : public FmtPass {
     }
 };
 
+/// Add newlines inside complex structures (arrays, objects etc.).
+///
+/// The main principle is that a structure can either be:
+/// * expanded and contain newlines in all the designated places
+/// * unexpanded and not contain any newlines in any of designated places.
+///
+/// It only looks shallowly at the AST nodes, so there may be some newlines deeper that
+/// don't affect expanding. For example:
+/// [{
+///     'a': 'b',
+///     'c': 'd',
+/// }]
+/// The outer array can be unexpanded, because there are no newlines between
+class FixNewlines: public FmtPass {
+    using FmtPass::visit;
+
+    bool shouldExpand(const Array *array)
+    {
+        for (const auto &elem: array->elements) {
+            if (countNewlines(open_fodder(elem.expr)) > 0) {
+                return true;
+            }
+        }
+        if (countNewlines(array->closeFodder) > 0) {
+            return true;
+        }
+        return false;
+    }
+
+    void expand(Array *array)
+    {
+        for (auto &elem: array->elements) {
+            ensureCleanNewline(open_fodder(elem.expr));
+        }
+        ensureCleanNewline(array->closeFodder);
+    }
+
+    Fodder &objectFieldOpenFodder(ObjectField &field)
+    {
+        if (field.kind == ObjectField::Kind::FIELD_STR) {
+            return field.expr1->openFodder;
+        }
+        return field.fodder1;
+    }
+
+    bool shouldExpand(Object *object)
+    {
+        for (auto &field: object->fields) {
+            if (countNewlines(objectFieldOpenFodder(field)) > 0) {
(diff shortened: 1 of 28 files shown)


sbarzowski:
Some pretty big test cases with a lot of examples are committed. Please take a look at them to check if the output style is acceptable.

The code in FixNewlines may look a little bit repetitive, but it's hard to avoid that. A very simple task is performed on many AST types. Some of it could be abstracted away, e. g.  by creating a function that returns "interesting" fodder objects for each. But I think we would still end up with similar number of functions and much less obvious code that requires some dancing around special cases.

There are also some improvements to test scripts:
* I got frustrated looking for a one line difference in 300 lines tests, so now diff is also presented
* A few times I passed .fmt.golden to `refresh_fmt_test`, so we check for that now (and to `refresh_test` as well).
* Now a temporary directory is created for running tests. It's only used to help with diff now, but it will be also useful for other tests in the future (for example `jsonnet fmt --in-place` tests).
* Now it's possible to set `VERBOSE` mode through ENV variable and not only by changing the code.

@lwander @sparkprime

test_suite/tests.source
```diff
@@ -18,9 +18,36 @@
...
+VERBOSE="${VERBOSE:-false}"
 
+separator() {
+    echo -e "\n-----------------------------\n"
+}
+
+get_temp_dir() {
+    if [ -z "$TMP_DIR" ]; then
+        # Simplest and most portable way I could find.
+        TMP_DIR="`python2 -c "import tempfile; print(tempfile.mkdtemp(prefix='jsonnet_'))"`"
```
lwander: Did you consider something along the lines of

```
TMP_DIR=$(mktemp -d jsonnet_XXXX)
```

core/formatter.cpp
```diff
@@ -963,6 +963,280 @@ class PrettyFieldNames : public FmtPass {
...
+            }
+            first = false;
+        }
+    }
+
+    // Example:
+    // very_long_function_on_a_long_line_so_arguments_dont_fit_on_the_same_line(
+    //     1, 2, 3)
+    // Should be expanded to:
+    // very_long_function_on_a_long_line_so_arguments_dont_fit_on_the_same_line(
```
lwander: I'm not familiar with the other formatting passes, but are any considerations taken for line length? Reading the code here it seems the expansion happens when args are already on another line, not if function is too long.

core/formatter.cpp
```diff
@@ -963,6 +963,280 @@ class PrettyFieldNames : public FmtPass {
...
+        }
+    }
+
+    // Example:
+    // f(1, 2,
+    //   3)
+    // Should be expanded to:
+    // f(1,
+    //   2,
+    //   3)
```
lwander: Shouldn't the closing paren live on a new line to match the [below style](https://github.com/google/jsonnet/pull/334/files#diff-7171b03fc4ff999fe62969d7161eba37R1168)? Or is that handled separately?

test_suite/tests.source
sbarzowski: Yes. There are portability problems. OS X and GNU versions of `mktemp` need different arguments. Invoking Python seems much cleaner to me than trying to autodetect the version or trying the other version if it fails (the latter is the top answer here https://unix.stackexchange.com/questions/30091/fix-or-alternative-for-mktemp-in-os-x).

core/formatter.cpp
sbarzowski: The point of this comment was to explain what this particular function does, expandBetween as opposed to expandNearParens (both operate on ArgParams). 

But still no, it wouldn't be touched by the other one (it would if there was a newline before 1 or after 3).

core/formatter.cpp
sbarzowski: No, it doesn't take the length into account. The name was only an example situation when someone would format it this way (as opposed to keeping arguments on the same line). Hmmm... but I can see how it can be confusing. Maybe it's better to change it to something generic.

core/formatter.cpp
```diff
@@ -963,6 +963,280 @@ class PrettyFieldNames : public FmtPass {
...
+/// * expanded and contain newlines in all the designated places
+/// * unexpanded and not contain any newlines in any of designated places.
+///
+/// It only looks shallowly at the AST nodes, so there may be some newlines deeper that
+/// don't affect expanding. For example:
+/// [{
+///     'a': 'b',
+///     'c': 'd',
+/// }]
+/// The outer array can be unexpanded, because there are no newlines between
```
sparkprime: Incomplete sentence

core/formatter.cpp
```diff
@@ -963,6 +963,280 @@ class PrettyFieldNames : public FmtPass {
     }
 };
 
+/// Add newlines inside complex structures (arrays, objects etc.).
+///
+/// The main principle is that a structure can either be:
+/// * expanded and contain newlines in all the designated places
+/// * unexpanded and not contain any newlines in any of designated places.
```
sparkprime: Slightly more concise description:

/// * unexpanded and contain newlines in none of the designated places.

sbarzowski: done

Issue #330, opened 2017-07-25
python 3 support?


lakshminagaraj:
Hello,

is python 3 support in pipeline? I see that currently on python 2.x is supported.

ant31:
duplicated https://github.com/google/jsonnet/issues/26

PR #335, opened 2017-08-07
Python 3 compatibility

### LINKED ISSUE #26: Python 3.x Compatibility
I forked and converted to use Visual Studio, and I also changed support from Python 2 C API to Python 3.4 API. It wasn't horribly difficult to do either task, but the project is there for anybody who needs or wants it.

benley:
Fixes #26, #330

### DIFF
--- a/python/_jsonnet.c
+++ b/python/_jsonnet.c
@@ -33,7 +33,11 @@ static const char *exc_to_str(void)
     PyObject *ptype, *pvalue, *ptraceback;
     PyErr_Fetch(&ptype, &pvalue, &ptraceback);
     PyObject *exc_str = PyObject_Str(pvalue);
+#if PY_MAJOR_VERSION >= 3
+    return PyUnicode_AsUTF8(exc_str);
+#else
     return PyString_AsString(exc_str);
+#endif
 }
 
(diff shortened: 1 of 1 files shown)


benley:
The one part I'm not sure about in this change is the `m_size` field in the `PyModuleDef _jsonnet` struct.  Does this module keep internal state in a way that requires an allocation here?

sparkprime:
m_size == -1 should be OK because all of the state is built up and torn down within the single evaluate call.  

Thanks for doing this!

ant31:
is it merged ?

benley:
I was wondering that too - maybe the merge happened in piper and we're waiting for MOE or copybara (or something) to run and publish the result back to github?

sparkprime:
I'm afraid the explanation was more simple than that :)  Sorry

benley:
No worries - thanks for merging :-D

wiseman:
Thanks for doing this work!

Is there an ETA for a new pypi release with support for Python 3?

sparkprime:
No plan just yet, if you need it make a bug and we'll figure out if there's anything we should wait for vs just cut it then and there.

PR #336, opened 2017-08-07
External testing support

sbarzowski:
Make it easier to use the test suite with other implementatons.
For example: https://github.com/google/go-jsonnet

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -91,16 +91,8 @@ default: jsonnet
 
 all: $(ALL)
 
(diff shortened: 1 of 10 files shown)


sbarzowski:
In addition to making it possible to use specified jsonnet binary this change removes some duplication and adds provides summary-only mode.
This change is required for this to make sense: https://github.com/google/go-jsonnet/pull/18

@lwander @sparkprime

sparkprime:
LGTM

PR #337, opened 2017-08-08
An option to disable error tests

sbarzowski:
Useful for testing other implementations which may have different error formats.


sbarzowski:
One more useful option for external testing.

I wonder how we are going to test that in the future. Aiming for a full compatibility of error messages doesn't sound like a very good option...

sparkprime:
We should test that the stack traces are correct (correct number of lines, correct file and source number).  As for the error messages, I don't see any reason why they should be different.  If they can be improved from the C++ version, we should change them in both.

As a last resort we can have separate golden files.  For example if you can improve the stack traces in a way that is very hard to do in the C++ version, we could do that.

PR #338, opened 2017-08-08
Add missing argument to the slice spec

sbarzowski:


sbarzowski:
I have noticed that when working on a Go version. @sparkprime

PR #341, opened 2017-08-14
Spec no-longer allows { f() +: e }

sparkprime:
@sbarzowski


sparkprime:
There was no desugaring rule for this construct, so it was confusing.

PR #342, opened 2017-08-14
Fix #172

### LINKED ISSUE #172: string in import statement is not escaped like regular strings
The spec is a bit vague on import, but the BNF clearly says it's a regular string, so it should process escape chars in the same way.

This change could cause grief for people on windows doing import "c:\blah.jsonnet".

On the other hand, the ability to import unicode files with \uXXXX codepoints may be useful.

There is another difference that also does not appear in the spec:  You cannot use ||| text blocks in import statements.  Either relax that restriction or update the spec.

sparkprime:
@sbarzowski

### DIFF
--- a/doc/language/spec.html
+++ b/doc/language/spec.html
@@ -138,7 +138,8 @@ The next line must be prefixed with some non-zero length whitespace <i>W</i>.  T
 first subsequent line that does not begin with <i>W</i>, and it is an error if this line does not
 contain some optional whitespace followed by <code>|||</code>.  The content of the string is the
 concatenation of all the lines that began with <i>W</i> but with that prefix stripped.  The line
-ending style in the file is preserved in the string.</li>
+ending style in the file is preserved in the string.  This form cannot be used in
+<code>import</code> statements.</li>
 </ul>
 
 <p>Double- and single-quoted strings are allowed to span multiple lines, in which case whatever
--- a/core/desugarer.cpp
+++ b/core/desugarer.cpp
@@ -643,11 +643,17 @@ class Desugarer {
             desugar(ast->body, obj_level);
             desugarParams(ast->params, obj_level);
 
-        } else if (dynamic_cast<const Import *>(ast_)) {
-            // Nothing to do.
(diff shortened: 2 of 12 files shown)


doc/language/spec.html
```diff
@@ -138,7 +138,8 @@ <h2 id="lexing">Lexing</h2>
 first subsequent line that does not begin with <i>W</i>, and it is an error if this line does not
 contain some optional whitespace followed by <code>|||</code>.  The content of the string is the
 concatenation of all the lines that began with <i>W</i> but with that prefix stripped.  The line
-ending style in the file is preserved in the string.</li>
+ending style in the file is preserved in the string.  This form cannot be used in
```
sbarzowski: A test for that would be nice.

core/ast.h
```diff
@@ -432,21 +432,21 @@ struct Function : public AST {
     }
 };
 
-struct LiteralString;
-
 /** Represents import "file". */
 struct Import : public AST {
-    LiteralString *file;
-    Import(const LocationRange &lr, const Fodder &open_fodder, LiteralString *file)
+    // Must be a LiteralString.
```
sbarzowski: Then why not keep it *LiteralString?

sparkprime: Because the desugarer visitor allows editing the AST *&, so everything has to be an AST *.

sparkprime: True of CompilerPass too, actually.

doc/language/spec.html
sparkprime: Done.

core/ast.h
sbarzowski: well, yes, but you can handle that while keeping it a *LiteralString, instead of:
```
desugar(ast->file, obj_level);
```
you could:
```
AST *file = ast->file;
desugar(file, obj_level);
ast->file = dynamic_cast<LiteralString *>(ast_);
```
I think it's better, because:
a) It's more explicit - it clearly states that it needs to be LiteralString, but in this case desugarer could potentially put something else there.
b) In general a little bit nicer data structures and a little less nice code within some functions is better than the other way around, imo.

sbarzowski:
LGTM

PR #357, opened 2017-09-06
Use a more common spelling: Pre-Procesor -> preprocessor

sbarzowski:
https://en.wikipedia.org/wiki/C_preprocessor
http://www.dictionary.com/browse/preprocessor?s=t


sbarzowski:
@sparkprime

PR #364, opened 2017-09-13
Consistently use name "number" rather than "double"

sbarzowski:
In jsonnet there is a single number type which is
in fact IEEE 754 double. We refer to it as "number"
when using `std.type` and in specification. However
in error messages and in our code it was still called double.
This change fixes this inconsistency by changing all occurances
to "number".


sbarzowski:
@lwander @sparkprime

PR #367, opened 2017-09-14
Fix swapped indexes in spec (object-inherit)

sbarzowski:
q is the number of assertions on the right side
p is the number of fields that exist on left side and not on the right
side
r is the number of fields that exist on right side and not on the left
side


sbarzowski:
@sparkprime

sparkprime:
The f's need apostrophes too.

sparkprime:
As do the h's and the e's :)

sbarzowski:
done

Issue #380, opened 2017-10-02
STATIC ERROR: std.jsonnet:1:1: Unexpected end of file.


nnmrts:
Hey! I tried using the binary version of jsonnet. However, whatever I do, it gives me back this error:

`STATIC ERROR: std.jsonnet:1:1: Unexpected end of file.`

I read on the jsonnet webpage [here](http://jsonnet.org/docs/stdlib.html), that std or stdlib is kind of a utility functions object for jsonnet itself. The error message suggests that the file is empty. Unfortunately we don't know really, which file is meant, it could be the std.jsonnet or the std.jsonnet.h file. At first I thought it's the first one, but that was full with 42kb of data. However, the problem is, there is an error, as I just discovered. A specific unexpected token prevents the file from being built and so the std.jsonnet.h file is empty in the end.

Will create a pull request fixing this issue and reference it here. Sorry for even submitting this issue now that I already discovered what the actual issue is. Nevertheless, I think it will bring more clarity to my pull request.

sbarzowski:
Thanks for reporting the issue. Could you provide some more details?
- What version of jsonnet are you using to get the error (jsonnet --version). You mentioned _binary version_, but where it's from?
- The specific command that leads to the error (steps to reproduce)
- Can you try reproducing the issue on current master?

The `:` is there for a reason - it's a slice (like in Python). This syntax is supported only in relatively new versions of jsonnet (and not recognized by some 3rd party tools like e.g. vscode extension). 

Here is some more context in case you are interested:
The standard library is embedded in the final executable or library. This means that std.jsonnet is used only in the build process, to build std.jsonnet.h - which is basically an array of bytes from std.jsonnet (https://github.com/google/jsonnet/blob/master/core/desugarer.cpp#L74). When some code is executed, the standard library is taken from memory (i.e. the array I mentioned earlier) and attached to that code (https://github.com/google/jsonnet/blob/master/core/desugarer.cpp#L884), with some additional magic for builtins etc.

nnmrts:
Hey! Sorry for providing so few information. These are the commands and versions and all of that, sorted by their time of installation:

### Versions:
```
Windows: 10.0.15063
Powershell: 5.1.15063.608
Python: 2.7.14:84471935ed
chocolatey: 0.10.8
Bazel: 0.6.0
jsonnet: 0.9.5
```

### Commands:
##### all executed in one session in the local jsonnet repo
```
set BAZEL_SH=C:\msys64\usr\bin\bash.exe
[...]

PR #381, opened 2017-10-02
removed unexpected token ":"

nnmrts:
fixing this issue: https://github.com/google/jsonnet/issues/380


sparkprime:
That is actually needed -- it says if you have a string like "-100" then take [1:] which is "100", convert that to a number, then make the number negative.

Let's figure out why having the : there broke your build?

sparkprime:
I see this PR predates the discussion in #380 but I'd like to get to the bottom of your problem

PR #383, opened 2017-10-03
Fix tests failing, but passing on Travis

sbarzowski:
We've shown the error message, but the exit code was still 0


sbarzowski:
@sparkprime

PR #387, opened 2017-10-12
Add --sort-imports to help, add --no-sort-imports

sbarzowski:


sbarzowski:
@sparkprime 

i've also added --no-sort-imports for consistency with other options.

PR #399, opened 2017-10-30
Add some Jsonnet programs that take a while to execute.

sparkprime:


perf_tests/large_string_join.jsonnet
```diff
@@ -0,0 +1,5493 @@
+{
+    text: std.join(',', [
```
sbarzowski: Can we avoid having thousands of lines here? Navigating this is pretty annoying...

sparkprime: you mean do `std.join(',', std.makeArray(n, function() 'x'))` ?

sure although it'd be interesting to see how performance differs between the two.

sparkprime: you made some benchmarks as well, right?

I'd like to have realistic ones if possible.

sbarzowski: Yeah, something like that.

For truly realistic examples a corpus of real code would be the best, probably.

sparkprime:
The one that golang used up all my RAM actually takes 30% of the RAM in C++ so could be explained by a constant factor on the golang side.  A constant factor of at least 2 is expected in garbage collectors, so it's not unexpected.

PR #401, opened 2017-10-31
Some UX improvements to commandline tool + cmdline tests

sparkprime:


test_cmd/run_cmd_tests.sh
```diff
@@ -0,0 +1,210 @@
...
+rm -rf out
+
+# And now the tests:
+
+# do_test <name> <exit_code> <args...>
+
+# Only check_file additional files if the test otherwise succeded, to avoid double-counting
+# failures.  All tests must write into out/$NAME/ and nowhere else.
+
+do_test "no_args" 1
```
sbarzowski: Can we extract the `do_test` part to a separate file? It will be easier to inspect them and add new ones without all of the machinery around them.

PR #416, opened 2017-12-03
Use directly static_cast, rather than probing with dynamic_cast

malkia:
Instead of probing and dynamic_cast-ing one, by one, use a switch over the stored AST node type, and then static_cast. This change speeds up ~2x times the evalution of  
https://github.com/ksonnet/ksonnet-lib/blob/master/examples/readme/hello-nginx.jsonnet

In addition:
- Added AST_APPLY_BRACE, as previously it was piggybacking on AST_BINARY
- Changed ASTs from std::vector<> to std::list<> as elements are not read by index, but one by one, and avoid reallocations. fwd_list might be even a better choice.

### DIFF
--- a/core/ast.h
+++ b/core/ast.h
@@ -30,6 +30,7 @@ limitations under the License.
 
 enum ASTType {
     AST_APPLY,
+    AST_APPLY_BRACE,
     AST_ARRAY,
     AST_ARRAY_COMPREHENSION,
     AST_ARRAY_COMPREHENSION_SIMPLE,
@@ -64,6 +65,7 @@ static inline std::string ASTTypeToString(ASTType type)
 {
     switch (type) {
         case AST_APPLY: return "AST_APPLY";
+        case AST_APPLY_BRACE: return "AST_APPLY_BRACE";
         case AST_ARRAY: return "AST_ARRAY";
         case AST_ARRAY_COMPREHENSION: return "AST_ARRAY_COMPREHENSION";
         case AST_ARRAY_COMPREHENSION_SIMPLE: return "AST_ARRAY_COMPREHENSION_SIMPLE";
@@ -126,7 +128,8 @@ struct AST {
     virtual ~AST(void) {}
 };
 
-typedef std::vector<AST *> ASTs;
+#include <list>
+typedef std::list<AST *> ASTs;
 
 /** Either an arg in a function apply, or a param in a closure / other function definition.
  *
@@ -214,7 +217,7 @@ struct ApplyBrace : public AST {
     AST *left;
     AST *right;  // This is always an object or object comprehension.
     ApplyBrace(const LocationRange &lr, const Fodder &open_fodder, AST *left, AST *right)
-        : AST(lr, AST_BINARY, open_fodder), left(left), right(right)
+        : AST(lr, AST_APPLY_BRACE, open_fodder), left(left), right(right)
     {
     }
 };
--- a/core/ast.h
+++ b/core/ast.h
@@ -21,6 +21,7 @@ limitations under the License.
 #include <cstdlib>
 
 #include <iostream>
+#include <list>
 #include <map>
 #include <string>
 #include <vector>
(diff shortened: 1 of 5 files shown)


sparkprime:
Thanks, I'm about to take a look.  Please can you post your actual performance numbers (e.g. 10s -> 5s)

core/pass.cpp
```diff
@@ -282,70 +282,50 @@ void CompilerPass::visit(Unary *ast)
...
-
-    } else {
-        std::cerr << "INTERNAL ERROR: Unknown AST: " << ast_ << std::endl;
-        std::abort();
+    switch(ast_->type) {
+        VISIT(ast_, AST_APPLY, Apply);
+        VISIT(ast_, AST_APPLY_BRACE, ApplyBrace);
+        VISIT(ast_, AST_ARRAY, Array);
+        VISIT(ast_, AST_ARRAY_COMPREHENSION, ArrayComprehension);
+        // VISIT(ast_, AST_ARRAY_COMPREHENSION, ArrayComprehensionSimple);
```
sbarzowski: What's happening here?

sparkprime:
You'll need to sign the CLA too.  This is good for all Google projects :)

core/pass.cpp
sparkprime: I think all these passes run before desugaring, so there should not be any ArrayComprehensionSimple.  That could change in future of course.

sparkprime:
@malkia  Just making sure you saw my note about the CLA

malkia:
Hi Dave! I'll take a look this weekend

malkia:
I signed it!

core/ast.h
```diff
@@ -126,7 +128,8 @@ struct AST {
     virtual ~AST(void) {}
 };
 
-typedef std::vector<AST *> ASTs;
+#include <list>
+typedef std::list<AST *> ASTs;
```
marcelocantos: I'm skeptical that `std::list` will improve anything. It requires more mallocs and consumes more memory than `std::vector`.

malkia:
Yes, it's not as critical as the dynamic_cast to static_cast with direct
switch statement. I need to profile it again, but can go without it.

Would remove it on my end, when I go back home.

On Dec 11, 2017 10:53, "Marcelo Cantos" <notifications@github.com> wrote:

*@marcelocantos* commented on this pull request.
------------------------------

In core/ast.h
<https://github.com/google/jsonnet/pull/416#discussion_r156164780>:

> @@ -126,7 +128,8 @@ struct AST {
     virtual ~AST(void) {}
 };

-typedef std::vector<AST *> ASTs;
+#include <list>
+typedef std::list<AST *> ASTs;

I'm skeptical that std::list will improve anything. It requires more
mallocs and consumes more memory than std::vector.

—
You are receiving this because you were mentioned.
Reply to this email directly, view it on GitHub
<https://github.com/google/jsonnet/pull/416#pullrequestreview-82598385>,
or mute
the thread

sparkprime:
I just tried it with std::list and std::vector on hello-nginx.jsonnet and it made no difference.

Also, that file was not particularly slow, at about 0.5s.  I'd love to see a real example that took > 10s as so far I only have synthetic examples that do that.

malkia:
Sorry, I'm failing to find time to work on this - but here is an example:

https://github.com/ksonnet/ksonnet-lib/blob/master/examples/readme/hello-nginx.jsonnet

This takes 3 secs using dynamic_cast and if/else blocks,
otherwise it takes 1.6secs with static_cast, and swtich/blocks.

Tested on Chromebook under crouton (Acer - not really powerful CPU there).

sparkprime:
I mean I tried it with static_cast and std::vector and I also tried it with static_cast and std::list and it was the same.  Obviously static_cast is a huge improvement over dynamic_cast.  I'll just merge it and change the list back to a vector.

sparkprime:
Thanks for the numbers by the way, it's interesting how much slower it is on Chromebook.

malkia:
Thanks Dave for taking care of that! Much appreciated I certainly not able to find much time to deal with it right now. Cheers, and Happy Holidays!

PR #417, opened 2017-12-05
Add JSONNET_PATH environment variable.

### LINKED ISSUE #411: FR: Allow configuring library search path from environment variable
It would be more convenient to set jsonnet library search path from environment variable instead of from argument.

```
export JSONNETPATH="${HOME}/abc"
jsonnet "$@"
# ==> equivalent to
# jsonnet -J "${JSONNETPATH}" "$@"
```

sparkprime:
Fix #411 (for cpp only)


sparkprime:
I didn't factor it out but I did make it about half as complex so I think it's fine now.

cmd/jsonnet.cpp
```diff
@@ -707,17 +708,11 @@ int main(int argc, const char **argv)
...
-                const char *path = strtok(input, const_cast<char*>(PATH_SEP));
-                input = nullptr;
-                if (path != nullptr) {
-                    jpath.push_front(path);
-                } else {
-                    break;
-                }
-            } while (true);
-            free(input);
+            std::stringstream ss(jsonnet_path_env);
```
marcelocantos: Use `std::ostringstream oss(…);` instead.

cmd/jsonnet.cpp
```diff
@@ -707,17 +708,11 @@ int main(int argc, const char **argv)
...
+            std::string path;
+            while (std::getline(ss, path, PATH_SEP)) {
+                jpath.push_front(path);
+            }
             for (const std::string &path : jpath) {
```
marcelocantos: Fold the two loops into one for good measure.

cmd/jsonnet.cpp
```diff
@@ -707,17 +708,11 @@ int main(int argc, const char **argv)
         const char *jsonnet_path_env = getenv("JSONNET_PATH");
         if (jsonnet_path_env != nullptr) {
```
marcelocantos: Suggest: `if (const char *jsonnet_path_env = getenv("JSONNET_PATH") {`

cmd/jsonnet.cpp
sparkprime: It's like that to reverse the order.

cmd/jsonnet.cpp
sparkprime: actually `istringstream` but yes, good point :)

PR #443, opened 2018-01-18
Chomp whitespace after |||, including \r, fix #421

### LINKED ISSUE #421: text-block |||  fails to allow trailing whitespace before new-line
From the spec, carriage returns are considered white space.

From the spec:
> Text block, beginning with |||, followed by **optional whitespace** and a new-line. The next line must be prefixed with some non-zero length whitespace W. The block ends at the first subsequent line that does not begin with W, and it is an error if this line does not contain some optional whitespace followed by |||. The content of the string is the concatenation of all the lines that began with W but with that prefix stripped. The line ending style in the file is preserved in the string. This form cannot be used in import statements.

Critically, the `|||` may be followed by an optional whitespace and then a new-line. As defined by the spec this would allow jsonnet to work with windows or unix line ending styles. However in implementation, both jsonnet and go-jsonnet both require a "\n" right after the "|||". This causes jsonnet files to fail with windows line endings if it contains a "|||" text block.

Other jsonnet features work just fine with CRLF line endings, including mutliline "normal" strings.

(This was discovered by a coworker who had git's auto crlf "feature" enabled.)

sparkprime:


sbarzowski:
I see that it preserves "/r" as they are in the block. Any reason not to just strip all carriage returns like discussed in the issue? That could be implemented as a one-liner `std::remove(input.begin(), input.end(), '\r')` before lexing.

sparkprime:
Yeah I thought about doing that but I figured it best to match the spec first.  The canonicalization of line endings is a change to the advertised behavior of the language

PR #449, opened 2018-01-25
Add std.mapWithIndex and std.mapWithKey

andreabedini:
`mapWithIndex` and `mapWithKey` are two functions that I always find useful. I am aware of thelist and object comprehension syntax but I tend to use it only for very short fragments of code. Naming is vaguely inspired by Haskell's `Data.Map`.

### DIFF
--- a/doc/docs/stdlib.html
+++ b/doc/docs/stdlib.html
@@ -65,6 +65,14 @@ hidden fields.</p>
 <p>Recursively remove all "empty" members of `x`.  "Empty" is defined as zero
 length `arrays`, zero length `objects`, or `null` values.</p>
 
+
+<h4 id="mapWithKey">std.mapWithKey(func, arr)</h4>
+
+<p>Apply the given function to all fields of the given object, also passing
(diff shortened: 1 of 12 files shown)


andreabedini:
I signed it!

sparkprime:
This looks useful, thanks.  Can you regenerate some of the golden test outputs to account for changed line numbers in stack traces that happen to go through std.jsonnet ?

What you do is:
cd test_suite
./refresh_golden.sh a.jsonnet b.jsonnet ...
until make test (in root of repo) works OK.

thanks

sparkprime:
Also can you add corresponding docs, see https://github.com/google/jsonnet/pull/450/files for an example

andreabedini:
> This looks useful, thanks. Can you regenerate some of the golden test outputs to account for changed line numbers in stack traces that happen to go through std.jsonnet ?

ah! that was the problem, sorry I had little time to investigate before.

> Also can you add corresponding docs, see https://github.com/google/jsonnet/pull/450/files for an example

Will do.

sparkprime:
Thanks

PR #450, opened 2018-01-25
Add missing functions to doc/docs/stdlib.html

gburiola:
std.isArray
std.isBoolean
std.isFunction
std.isNumber
std.isObject
std.isString
std.log
std.manifestJsonEx
std.setMember
std.strReplace

Some of these functions were added after last release `v0.9.5` so probably better to only merge this MR when we're close to cut a new release.


sparkprime:
Thanks for helping with the documentation

sparkprime:
I think it's OK if the website refers to HEAD.  We should do a release soon anyway

gburiola:
How long does it take for the website to sync with the repo ?

PR #454, opened 2018-01-30
update function std.strReplace

gburiola:
If `from_len==1` it's much faster to split / rejoin the string

### lib.jsonnet

```
{
    strReplace(str, from, to)::
        assert std.type(str) == "string";
        assert std.type(from) == "string";
        assert std.type(to) == "string";
        assert from != "" : "'from' string must not be zero length.";

        // Cache for performance.
        local str_len = std.length(str);
        local from_len = std.length(from);

        // True if from is at str[i].
        local found_at(i) = str[i:i + from_len] == from;

        // Return the remainder of 'str' starting with 'start_index' where
        // all occurrences of 'from' after 'curr_index' are replaced with 'to'.
        local replace_after(start_index, curr_index, acc) =
            if curr_index > str_len then
                acc + str[start_index:curr_index]
            else if found_at(curr_index) then
                local new_index = curr_index + std.length(from);
                replace_after(new_index, new_index, acc + str[start_index:curr_index] + to)
            else
                replace_after(start_index, curr_index + 1, acc);

        // if from_len==1, then we replace by splitting and rejoining the
        // string which is much faster than replace_after
        if from_len == 1 then
            std.join(to, std.split(str, from))
        else
            replace_after(0, 0, ""),
}
```

### baseline

```
local lib = import "lib.jsonnet";

{
  values: {
    [std.toString(i)]: "www.google.com" for i in std.range(100,999)
  }
}
```

```
$ (time jsonnet baseline.jsonnet) 2>&1 | egrep "999|real"
      "999": "www.google.com"
real    0m0.130s
```

### strReplace (with from_len==1)

```
local lib = import "lib.jsonnet";
{
  values: {
    [std.toString(i)]: lib.strReplace("www.google.com", ".", "__") for i in std.range(100, 999)
  },
}
```

```
$ (time jsonnet strReplace.jsonnet) 2>&1 | egrep "999|real"
      "999": "www__google__com"
real	0m1.974s
```

### strReplace (with from_len>1)

```
local lib = import "lib.jsonnet";
{
  values: {
    [std.toString(i)]: lib.strReplace("www.google.com", "com", "org") for i in std.range(100, 999)
  },
}
```

```
$ (time jsonnet strReplace2.jsonnet) 2>&1 | egrep "999|real"
      "999": "www.google.org"
real	0m11.954s
```

### DIFF
--- a/stdlib/std.jsonnet
+++ b/stdlib/std.jsonnet
@@ -129,7 +129,13 @@ limitations under the License.
             else
                 replace_after(start_index, curr_index + 1, acc);
 
-        replace_after(0, 0, ""),
+        // if from_len==1, then we replace by splitting and rejoining the
+        // string which is much faster than recursing on replace_after
+        if from_len == 1 then
+            std.join(to, std.split(str, from))
+        else
+            replace_after(0, 0, ""),
+
 
     range(from, to)::
         std.makeArray(to - from + 1, function(i) i + from),
--- a/test_suite/error.equality_function.jsonnet.golden
+++ b/test_suite/error.equality_function.jsonnet.golden
@@ -1,3 +1,3 @@
 RUNTIME ERROR: Cannot test equality of functions
-	std.jsonnet:1063:17-42	function <anonymous>
+	std.jsonnet:1069:17-42	function <anonymous>
 	error.equality_function.jsonnet:17:1-33	
--- a/test_suite/error.inside_equals_array.jsonnet.golden
+++ b/test_suite/error.inside_equals_array.jsonnet.golden
@@ -1,7 +1,7 @@
 RUNTIME ERROR: foobar
 	error.inside_equals_array.jsonnet:18:18-32	thunk <array_element>
-	std.jsonnet:1043:41-45	thunk <b>
-	std.jsonnet:1043:33-45	function <anonymous>
-	std.jsonnet:1043:33-45	function <aux>
-	std.jsonnet:1046:29-45	function <anonymous>
-	std.jsonnet:1047:21-33	
+	std.jsonnet:1049:41-45	thunk <b>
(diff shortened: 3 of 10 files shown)


doc/docs/stdlib.html
```diff
@@ -182,6 +182,17 @@ <h4 id="strReplace">std.strReplace(str, from, to)</h4>
    yields <code>"I like to surf with my surfboard"</code>. </p>
 
 
+<h4 id="charReplace">std.charReplace(input_str, from_char, to_str)</h4>
+
+<p>Returns a copy of the string in which all occurrences of char <code>from_char</code> have been replaced
+   with string <code>to_str</code></p>
+
+<p>This is ~6x faster than <code>std.strReplace</code> when replacing single characters</p>
```
sbarzowski: Nit: I would just write that it's "much" faster, not 6x faster. The difference in speed depends on specific benchmark and it will be a pain to maintain as the implementation evolves.

sbarzowski:
BTW current strReplace implementation is pretty naive it could be much faster.

doc/docs/stdlib.html
gburiola: updated. thanks

sbarzowski:
LGTM

@sparkprime

sparkprime:
How about using a conditional to optimize the strReplace case when the `from` argument has length 1?

gburiola:
I've upgraded this with the suggestion from @sparkprime 

However, I'll close this PR in favor of https://github.com/google/go-jsonnet/pull/186 witch has a much better native implementation.

sbarzowski:
This is still It's still an improvement for C++ jsonnet, so I would merge it.

sparkprime:
thanks

PR #457, opened 2018-02-15
Fix mapWithKey parameter type in documentation

roidelapluie:
Signed-off-by: Julien Pivotto <roidelapluie@inuits.eu>

### DIFF
--- a/doc/docs/stdlib.html
+++ b/doc/docs/stdlib.html
@@ -74,7 +74,7 @@ hidden fields.</p>
 length `arrays`, zero length `objects`, or `null` values.</p>
 
 
-<h4 id="mapWithKey">std.mapWithKey(func, arr)</h4>
+<h4 id="mapWithKey">std.mapWithKey(func, obj)</h4>
 
 <p>Apply the given function to all fields of the given object, also passing
   the field name. The function <code>func</code> is expected to take the


roidelapluie:
I signed it!

doc/docs/stdlib.html
```diff
@@ -74,7 +74,7 @@ <h4 id="prune">std.prune(x)</h4>
 length `arrays`, zero length `objects`, or `null` values.</p>
 
 
-<h4 id="mapWithKey">std.mapWithKey(func, arr)</h4>
+<h4 id="mapWithKey">std.mapWithKey(func, o)</h4>
```
sparkprime: It's actually `obj` not `o`

This actually matters because of the named parameter passing:

```
dcunnin@dcunnin7:~/jsonnet$ jsonnet -e 'std.mapWithKey(function(a, b)[a,b], o={})'
RUNTIME ERROR: Function has no parameter o
```

andreabedini:
Oopsie, thanks for fixing this!

doc/docs/stdlib.html
roidelapluie: Then this is inconsistent with the other functions that have objects as parameters?

sparkprime: Fair point. I guess we should update the jsonnet file to rename obj to o then. The docs have to reflect the actual code (and hopefully one day will be generated from the code). Thanks for noticing this.

sparkprime: On the other hand these single letter parameters date from before we had named parameter passing, therefore they basically were exposed by accident and aren't the ideal param names.  I think I'd prefer if they all had names like `obj`.  I'd understand if that's not what you signed up for though :)

sparkprime: Ok I reviewed the whole stdlib - there are a great many places where the parameters are inconsistent / too terse.  I should have fixed them all when I introduced the named parameter passing.  I don't think fixing them all now is a good idea because it's potentially a breaking change at least for some of them.  So I propose leaving them as they are for now, and doing a big overhaul of the stdlib in the next few months to:

- Add a ton more needed functions
- Divide it into independent libraries
- Fix some issues like parameter ordering, naming of functions, and the parameter names we're discussing now.

The design of the stdlib has been an enduring embarassment. Community participation in its replacement would be greatly appreciated!

So if you could please update this doc element to match the existing code (`obj`), that's the right thing to do now.

roidelapluie: OK. Updated.

roidelapluie: (The real improvement would be auto-generated doc.)

Issue #359, opened 2017-09-07
Better defaults for jsonnet fmt


sbarzowski:
Right now `jsonnet fmt` by default does less than I would expect, e.g. it doesn't fix indentation. 

Something like `jsonnet fmt -n 2 --max-blank-lines 1 --sort-imports` could be the default. 

I think that ideally long-term we would have something like `gofmt` which provides heavily opinionated, "the only true" formatting, without any knobs. Having reasonable defaults is the first step in this direction.

sparkprime:
My vote would be for:

`-n 2 --max-blank-lines 2 --sort-imports --string-style s --comment-style s`

This is what we now use internally, in the system you helped implement when you were here :)

PR #462, opened 2018-02-23
Change default formatter params

sparkprime:


sparkprime:
I thought it made sense to pick one and that one is a bit visually cleaner.  It could have gone either way though.  I don't think PEP8 or the Google Python style guide expresses a preference.

sparkprime:
Fixes #359

PR #464, opened 2018-02-24
Use Mac-friendly ANSI-escapes in test output

marcelocantos:
(Also include jsonnet cmd in output messages.)

macOS's `echo` is behind the times and doesn't support ANSI escapes. Its `printf` does, but only via `\033`, not `\e`.


sbarzowski:
Great! I was wondering why it didn't work on my Mac.

LGTM @sparkprime

sparkprime:
Good to know, thanks

PR #465, opened 2018-02-24
Replace "hermiticity" with "hermeticity" in docs

marcelocantos:
… on the assumption that the intended meaning was _the condition of being hermetic_, not _the condition of being Hermitian_.


sparkprime:
:) thanks

PR #466, opened 2018-02-25
Several stdlib changes

### LINKED ISSUE #453: Add upper/lower case functions to stdlib
It would be useful to have functions in stdlib to change the case of a string to all upper or all lower.

### LINKED ISSUE #461: std.manifestYAML
It would be handy to format output as "canonical" YAML (i.e. given json data, use basic YAML features for readability).

Things people seem to care about:
* not emitting { } or [ ]
* using multiline block syntax for string when applicable
* keeping strings quoted?
* what else?

Is there an existing spec for canonical YAML?

It should be easy enough to modify `manifestJsonEx` to implement `manifestYaml`.

sparkprime:
- Fix std.manifestJson escaping in field attributes
- add YAML and XML manifestation, fixes #461
- add asciiUpper and asciiLower, fixes #453

### DIFF
--- a/doc/docs/stdlib.html
+++ b/doc/docs/stdlib.html
@@ -379,7 +379,8 @@ one or more whitespaces that are used for indentation:</p>
 <h4 id="manifestYamlDoc">std.manifestYamlDoc(value)</h4>
 
 <p>Convert the given value to a YAML form.  Note that <code>std.manifestJson</code> could also be
-used for this purpose, but this function will produce more canonical-looking YAML.</p>
+used for this purpose, because any JSON is also valid YAML.  But this function will produce more
+canonical-looking YAML.</p>
 
 <code>std.manifestYamlDoc(
 {
@@ -428,12 +429,12 @@ used for this purpose, but this function will produce more canonical-looking YAM
 </code>
 
 
-<h4 id="manifestXml">std.manifestXml(xml)</h4>
+<h4 id="manifestXmlJsonml">std.manifestXmlJsonml(value)</h4>
 
 <p>Convert the given <a href="http://www.jsonml.org/">JsonML</a>-encoded value to a string
 containing the XML.</p>
 
-<code>std.manifestXml([
+<code>std.manifestXmlJsonml([
   'svg',
   { height: 100, width: 100 },
   ['circle', { cx: 50, cy: 50, r: 40, stroke: 'black', 'stroke-width': 3, fill: 'red' }],
@@ -446,7 +447,7 @@ containing the XML.</p>
 &lt;/svg&gt;
 </code>
 
-<p>Which looks like this:</p>
+<p>Which represents the following image:</p>
 
 <svg height="100" width="100">
   <circle cx="50" cy="50" r="40" stroke="black" stroke-width="3" fill="red" />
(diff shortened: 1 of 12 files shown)


stdlib/std.jsonnet
```diff
@@ -136,6 +136,22 @@ limitations under the License.
...
+      std.char(cp(c) - 32)
+    else
+      c;
+    std.join('', std.map(up_letter, std.stringChars(x))),
+
+  asciiLower(x)::
+    local cp = std.codepoint;
+    local down_letter(c) = if cp(c) >= 65 && cp(c) < 91 then
+      std.char(cp(c) + 32)
+    else
```
sbarzowski: nit: the indentation looks a bit weird here, it looks as if `else` matched `local`

stdlib/std.jsonnet
```diff
@@ -136,6 +136,22 @@ limitations under the License.
     else
       replace_after(0, 0, ''),
 
+  asciiUpper(x)::
+    local cp = std.codepoint;
+    local up_letter(c) = if cp(c) >= 97 && cp(c) < 123 then
```
sbarzowski: nit: Magic numbers (but it's easy enough to figure out what they mean).

doc/docs/stdlib.html
```diff
@@ -360,6 +376,94 @@ <h4 id="manifestJsonEx">std.manifestJsonEx(value, indent)</h4>
...
+])</code>
+
+<p>Yields a string containing this XML (all on one line):</p>
+
+<code>&lt;svg height="100" width="100"&gt;
+  &lt;circle cx="50" cy="50" fill="red" r="40" stroke="black" stroke-width="3"&gt;&lt;/circle&gt;
+&lt;/svg&gt;
+</code>
+
+<p>Which looks like this:</p>
```
sbarzowski: Nit: technically it says that the string above looks like the circle below, while it only represents that circle.

doc/docs/stdlib.html
```diff
@@ -360,6 +376,94 @@ <h4 id="manifestJsonEx">std.manifestJsonEx(value, indent)</h4>
 </code>
 
 
+<h4 id="manifestYamlDoc">std.manifestYamlDoc(value)</h4>
+
+<p>Convert the given value to a YAML form.  Note that <code>std.manifestJson</code> could also be
+used for this purpose, but this function will produce more canonical-looking YAML.</p>
```
sbarzowski: Nit: may be a good place to mention that any json is a valid yaml

stdlib/std.jsonnet
```diff
@@ -864,19 +889,61 @@ limitations under the License.
...
+        local len = std.length(v);
+        if len == 0 then
+          '""'
+        else if v[len - 1] == '\n' then
+          local split = std.split(v, '\n');
+          std.join('\n' + cindent, ['|'] + split[0:std.length(split) - 1])
+        else
+          std.escapeStringJson(v)
+      else if std.type(v) == 'function' then
+        error 'Tried to manifest function at ' + path
```
sbarzowski: 👍

stdlib/std.jsonnet
sparkprime: It's what the formatter does, and I think it's ok because it's like

```
local x = [
  0,
];
```

stdlib/std.jsonnet
sparkprime: If we had try / catch we would add the context into those errors too

sparkprime:
I realized that this is only one way of encoding XML, and it's not even necessary the best one for most use cases.  Usually, you don't need the text between the elements.  Also, its use of arrays makes it hard to abstract it using Jsonnet mixins.

Issue #441, opened 2018-01-17
Any plan for a new release?


fajran:
I would like to use Bazel to call Jsonnet and produce a Yaml file of Json documents. I have [my changes](https://github.com/fajran/rules_jsonnet/commit/42d467a0f95cc96678031383def89f070c8a3335) ready for the Jsonnet Bazel rules but it requires a newer version of Jsonnet, newer than 0.9.5, or something that includes 381d7ddd.

If we have a new release, I will create a PR to [bazelbuild/rules_jsonnet](https://github.com/bazelbuild/rules_jsonnet) project to use the new version.

thanks!

sparkprime:
Yes it's on my TODO list.  I think it was a mistake to remove the deprecated commandline options in 93055441 though, they need to go through a phase of giving warning messages before we actually remove them.  I'll add them back in before I do a release.  (Bazel might actually be affected by that).

kinghrothgar:
Sorry to be pushy, but any update on when this might be released?

sbarzowski:
I think the commandline flags mentioned before were brought back here: https://github.com/google/jsonnet/commit/5f0a9fd7214b26420e8853f5f9d41ca6701972ae.

@sparkprime Are there any remaining blockers?

sparkprime:
No more blockers but release is kinda time consuming with needing to check every platform and write release notes.  We're going through "perf" at google right now so I will get to it hopefully later this week.

PR #472, opened 2018-03-15
Updated documentation with new defaults

kinghrothgar:
This updates the help docs to reflect the default changes in commit 0f1addef9137f2f27cd0d5dfdd33aaebba8efce9.

In reference to #441, I think these default changes may mean a "major" version bump is required.


sparkprime:
Nice, thanks

sparkprime:
You're probably right about the major version.  Although it is just the formatter, we don't have separate versioning for that and it is a big change for anyone who habitually formats their code (and chooses not to override the defaults).

PR #481, opened 2018-03-29
Keyed set funcs

redbaron:
WIP, untested. Just testing waters :) Is it something jsonnet devs would be interesting to have in stdlib?

### DIFF
--- a/stdlib/std.jsonnet
+++ b/stdlib/std.jsonnet
@@ -23,6 +23,7 @@ limitations under the License.
 {
 
   local std = self,
+  local id = function(x) x,
 
   isString(v):: std.type(v) == 'string',
   isNumber(v):: std.type(v) == 'number',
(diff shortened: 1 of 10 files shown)


sbarzowski:
Seems like a very natural addition. I was thinking about adding it myself. Especially since `<`  doesn't work on arrays (perhaps it should, btw) so the standard workaround is not available.

sparkprime:
I guess I have no objection.  I suppose this is useful for de-duping objects which have a name field with same value, but are otherwise different?

redbaron:
I am using these to replace k8s resources in Ksonnet component with newer and better versions and to check for duplicates post all customizations:

```
local resourceKeyF = function(r) "" + r.kind + r.metadata.name + r.metadata.namespace
...
+ {
   assert std.length(std.set(self.items, resourceKeyF)) == std.length(self.items) : "Duplicate resources found"
}
```

stdlib/std.jsonnet
```diff
@@ -1066,45 +1069,46 @@ limitations under the License.
...
       []
     else
-      local pivot = arr[0];
+      local pivot = keyF(arr[0]);
       local rest = std.makeArray(l - 1, function(i) arr[i + 1]);
-      local left = std.filter(function(x) x <= pivot, rest);
-      local right = std.filter(function(x) x > pivot, rest);
-      std.sort(left) + [pivot] + std.sort(right),
+      local left = std.filter(function(x) keyF(x) < pivot, rest);
+      local right = std.filter(function(x) keyF(x) >= pivot, rest);
```
redbaron: When merging sets, often you have preference which elements to keep, therefore sorting need to be stable, then values coming first are "winning". This change seems to me enough to make quicksort stable, but please have a look

sbarzowski: I think it works.

Proof:
It works for length 0. Not let's assume it's stable for all lengths < n (*). Filter doesn't change the relative order of elements < pivot. By (*) after sorting, the order of equal elements in this subarray is maintained. Similarly for elements >= pivot. Pivot is the first element of the array, so you know that it goes before any elements >= pivot.  So the order of equal elements is maintained.

redbaron:
Code in its current form is used in our projects, not for long but seems to be working fine.  There are some errors in travis-ci `INTERNAL ERROR: std.jsonnet not an object.` though, but I can't spot anything wrong

sbarzowski:
Travis failure looks like a configuration problem.  If always fails here: https://github.com/google/jsonnet/blob/master/core/desugarer.cpp#L889. I couldn't reproduce it on my machine.

sbarzowski:
BTW It would be nice if this change also included some tests and appropriate changes to the documentation (this file: https://github.com/google/jsonnet/blob/master/doc/docs/stdlib.html). Ideally it would include a note that this new parameter was added in version UNRELEASED (in the past people got confused because of functions that were not available in the version they were using).

redbaron:
@sbarzowski I think adding `local id = function` makes it upset somehow. I checked master and there is no `std.jsonnet is not an object` error, just adding `local id = true` to std file caused and error. I'll move local inside and see how it goes

stdlib/std.jsonnet
```diff
@@ -1066,64 +1067,65 @@ limitations under the License.
...
+        else if keyF(a[i]) < keyF(b[j]) then
           aux(a, b, i + 1, j, acc) tailstrict
         else
           aux(a, b, i, j + 1, acc) tailstrict;
     aux(a, b, 0, 0, []) tailstrict,
 
-  setDiff(a, b)::
-    local aux(a, b, i, j, acc) =
+  setDiff(a, b, keyF=id)::
+    local aux(a, b, i, j, acc, keyF) =
```
redbaron: probably it is some strange effect of `tailstrict`, which I don't understand, but `aux` didn't capture `keyF` arg from `setDiff`, producing unknown variable error. Strangely enough, it managed to capture `std` which is "further" from it

sparkprime: I tried reproducing this but it worked fine: redbaron#1

sbarzowski:
It passes the tests on Travis now, even though I'm not sure what has changed (appveyor part seems to fail for any PR).

redbaron:
I am wondering, should it be `keyF` at all? maybe `compareF` is better? or maybe have both?

sbarzowski:
`compareF` is more general, but `keyF` is nicer to use most of the time. One argument for `compareF` over `keyF` is that `<` doesn't work on arrays right now, so for example to sort by last name, then by first name, you need to sort twice or hack it somehow.

Issue #130, opened 2016-03-15
A trace function


copumpkin:
There might well be one already, but I looked under the "Assertions and Debugging" section of the stdlib docs and only saw assertions.

Basically, I want a function that is equivalent to

`local const(str, x) = x;`

But that prints `str` to `stderr` as a side effect when forced. Very handy for debugging odd execution flow and the like.

Both Haskell and Nix call this function `trace`, but I'm open to other ideas.

sparkprime:
Yes, this is a great idea and something I've had in the back of my mind for a while.  You can sort of debug with error right now -- error foo will print out foo, but will also block execution.

One concern is how many implementation hacks this would expose.  For example, at the moment, object-level locals are re-evaluated whenever you read a field of the object (as they are desugared into every field).  They are not cached like regular thunks.  This will give surprising output!  However it is probably still useful.

copumpkin:
Hah, didn't realize that. Yeah, I don't think anyone will be expecting particularly sensible properties around the output, other than it somehow representing what the VM is doing.

sparkprime:
It is trivial to add a new builtin that prints str and returns x.  I'm not saying that would be the best way to provide this feature, but it would give you something to play with if you wanted to do that.

copumpkin:
I've been meaning to poke around the code for a while. I'll see if I can scrounge together some time one of these days and hack at it, if you don't get sick of waiting for me before that :smile:

sparkprime:
To save some time -- you need to add another builtin to jsonnet_builtin_decl in desugarer.cpp then modify vm.cpp to add the corresponding case to the inner interpreter loop

                     case FRAME_BUILTIN_FORCE_THUNKS: {

and I apologize for the hairiness of that code :)

sbarzowski:
https://stackoverflow.com/questions/49742536/how-to-printf-in-jsonnet/49760216#49760216
The need for this keeps coming back. it's pretty clear people want to debug jsonnet sometimes :-).

ahakanbaba:
I will push a PR soon.

PR #487, opened 2018-04-10
Add a builtin trace function.

ahakanbaba:
Prints the first argument to std err with line number
Returns the second argument which is expected to be an object.

ISSUE: #130

### DIFF
--- a/core/desugarer.cpp
+++ b/core/desugarer.cpp
@@ -63,7 +63,7 @@ BuiltinDecl jsonnet_builtin_decl(unsigned long builtin)
         case 24: return {U"primitiveEquals", {U"a", U"b"}};
         case 25: return {U"native", {U"name"}};
         case 26: return {U"md5", {U"str"}};
-        case 27: return {U"trace", {U"str", U"obj"}};
+        case 27: return {U"trace", {U"str", U"rest"}};
         default:
             std::cerr << "INTERNAL ERROR: Unrecognized builtin function: " << builtin << std::endl;
             std::abort();
--- a/core/desugarer.cpp
+++ b/core/desugarer.cpp
@@ -32,7 +32,7 @@ struct BuiltinDecl {
     std::vector<UString> params;
 };
 
-static unsigned long max_builtin = 26;
+static unsigned long max_builtin = 27;
 BuiltinDecl jsonnet_builtin_decl(unsigned long builtin)
 {
     switch (builtin) {
@@ -63,6 +63,7 @@ BuiltinDecl jsonnet_builtin_decl(unsigned long builtin)
         case 24: return {U"primitiveEquals", {U"a", U"b"}};
         case 25: return {U"native", {U"name"}};
         case 26: return {U"md5", {U"str"}};
+        case 27: return {U"trace", {U"str", U"obj"}};
         default:
             std::cerr << "INTERNAL ERROR: Unrecognized builtin function: " << builtin << std::endl;
             std::abort();
--- a/core/state.h
+++ b/core/state.h
@@ -43,9 +43,7 @@ struct Value {
         ARRAY = 0x10,
         FUNCTION = 0x11,
         OBJECT = 0x12,
-        STRING = 0x13,
-
-        ANY = 0xFF00
+        STRING = 0x13
     };
     Type t;
     union {
@@ -70,7 +68,6 @@ std::string type_str(Value::Type t)
         case Value::FUNCTION: return "function";
         case Value::OBJECT: return "object";
         case Value::STRING: return "string";
-        case Value::ANY: return "valuetype";
         default:
             std::cerr << "INTERNAL ERROR: Unknown type: " << t << std::endl;
             std::abort();
--- a/core/state.h
+++ b/core/state.h
@@ -43,7 +43,9 @@ struct Value {
         ARRAY = 0x10,
         FUNCTION = 0x11,
         OBJECT = 0x12,
-        STRING = 0x13
+        STRING = 0x13,
+
+        ANY = 0xFF00
     };
     Type t;
     union {
@@ -68,6 +70,7 @@ std::string type_str(Value::Type t)
         case Value::FUNCTION: return "function";
         case Value::OBJECT: return "object";
         case Value::STRING: return "string";
+        case Value::ANY: return "valuetype";
         default:
             std::cerr << "INTERNAL ERROR: Unknown type: " << t << std::endl;
             std::abort();
--- a/doc/docs/stdlib.html
+++ b/doc/docs/stdlib.html
@@ -632,5 +632,5 @@ local conditionalReturn(cond, in1, in2) =
         std.trace('cond is true returning ' + std.toString(n1), in1)
     else
         std.trace('cond is false returning ' + std.toString(in2), in2);
-}</code></p>
+</code></p>
 
--- a/doc/docs/stdlib.html
+++ b/doc/docs/stdlib.html
@@ -622,15 +622,30 @@ href="https://tools.ietf.org/html/rfc7396">RFC7396</a></p>
 
 <h4 id="trace">std.trace(str, rest)</h4>
 
-<p> <code>trace</code> outputs the string to std::cerr in its first argument, before
-returning the second argument as its result </p>
+<p><code>trace</code> outputs the given string <code>str</code> to stderr and
+returns <code>rest</code> as the result.<p>
+<em>Available since version 0.11.0</em>
 
 <p>Example:<br />
 <code>
 local conditionalReturn(cond, in1, in2) =
     if (cond) then
-        std.trace('cond is true returning ' + std.toString(n1), in1)
+        std.trace('cond is true returning ' + std.toString(in1), in1)
     else
         std.trace('cond is false returning ' + std.toString(in2), in2);
+
+
+{
+    a: conditionalReturn(true, { b: true }, { c: false }),
+}
 </code></p>
 
+<p>Prints:</p>
+<p><code>
+TRACE: test.jsonnet:3 cond is true returning {"b": true}
+{
+   "a": {
+      "b": true
+   }
+}
+</code></p>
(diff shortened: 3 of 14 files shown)


ahakanbaba:
@sparkprime  
Added a trace function that (I think) is similar to the haskell's trace function 
From [here](https://wiki.haskell.org/Debugging): 

> "When called, trace outputs the string in its first argument, before returning the second argument as its result.'"

I have not added documentation or error tests yet. 

As it stands the trace function expects an obj `{}` as its second argument. I am not sure how useful that is. 

Maybe we should implement a trace function that can return whatever its second argument is ?

ahakanbaba:
By the way, are the appveyor tests operational ? 
I am getting this error that I do not think I have caused

```C:\Program Files (x86)\MSBuild\Microsoft.Cpp\v4.0\V140\Microsoft.Cpp.Platform.targets(57,5): error MSB8020: The build tools for v141 (Platform Toolset = 'v141') cannot be found. To build using the v141 build tools, please install v141 build tools.  Alternatively, you may upgrade to the current Visual Studio tools by selecting the Project menu or right-click the solution, and then selecting "Retarget solution". [C:\projects\jsonnet-dklrc\vs2017\core.vcxproj]```

**Comment by sbarzowski** [comment]
Don't worry about appveyor tests. There were some problems going on for a while.

**Comment by sbarzowski** [comment]
> As it stands the trace function expects an obj {} as its second argument. I am not sure how useful that is.
> 
> Maybe we should implement a trace function that can return whatever its second argument is ?

Yes, I think it should allow any value to be passed there. It's a bit confusing otherwise. Especially since it's intended for debugging, so the user may not know about the type of the value passed to it.

File: core/desugarer.cpp [hunk 1]
```diff
@@ -63,6 +63,7 @@ BuiltinDecl jsonnet_builtin_decl(unsigned long builtin)
         case 24: return {U"primitiveEquals", {U"a", U"b"}};
         case 25: return {U"native", {U"name"}};
         case 26: return {U"md5", {U"str"}};
+        case 27: return {U"trace", {U"str", U"obj"}};
```
ahakanbaba: Once the trace call can accept any type as the second parameter, what to call this here ? Right now it is obj. 
Does it actually matter except what is printed in an error message ?

core/state.h
```diff
@@ -43,7 +43,9 @@ struct Value {
         ARRAY = 0x10,
         FUNCTION = 0x11,
         OBJECT = 0x12,
-        STRING = 0x13
+        STRING = 0x13,
+
+        ANY = 0xFF00
```
ahakanbaba: I am not sure whether it is acceptable to add this ENUM. I am open to suggestions though.

ahakanbaba:
As @sparkprime  mentioned in #130 the trace reveals implementation details. 
One thing I have realized the trace execution is not necessarily the order of declaration in the code. 

For example 

```
// local innerrFunc = function(x,y)
local func() = {
    t: std.trace("1",{a:"b"}),
    k: std.trace("2",[{},{a:"b"}]),
    l: std.trace("3",15),
    m: std.trace("4","text"),
    n: std.trace("5",true),
    x: std.trace("6",[{},{a:"b"}]),
    r: std.trace("8",true),
    z: std.trace("9",[{},{a:"b"}]),
    s: std.trace("10",true),
    o: std.trace("11",null),
};

{
   otherKey: func(),
   key: std.foldl(std.trace("Hello",function(acc,i) acc + i), [1,2,3],0),
}```

Prints 

```
$ ~/src/google/jsonnet/jsonnet  test.jsonnet
TRACE: test.jsonnet:17 Hello
TRACE: test.jsonnet:4 2
TRACE: test.jsonnet:5 3
TRACE: test.jsonnet:6 4
TRACE: test.jsonnet:7 5
TRACE: test.jsonnet:12 11
TRACE: test.jsonnet:9 8
TRACE: test.jsonnet:11 10
TRACE: test.jsonnet:3 1
TRACE: test.jsonnet:8 6
TRACE: test.jsonnet:10 9
{
   "key": 6,
   "otherKey": {
      "k": [
         { },
         {
            "a": "b"
         }
      ],
      "l": 15,
      "m": "text",
      "n": true,
      "o": null,
      "r": true,
      "s": true,
      "t": {
         "a": "b"
      },
      "x": [
         { },
         {
            "a": "b"
         }
      ],
      "z": [
         { },
         {
            "a": "b"
         }
      ]
   }

```}

**Comment by ahakanbaba** [comment]
If all this looks reasonable I will also add documentation and then this can be merged. 
Please take a look

File: core/desugarer.cpp (hunk 1 above)
Comment (sparkprime) [reply core/desugarer.cpp]: It doesn't really matter.  `rest` might work.

Another option here is to return `true`, in which case this can be used as in if `std.trace("foo") then ...` or just `assert std.trace("foo"); ...`

Comment (ahakanbaba) [reply core/desugarer.cpp]: I do not understand what you suggest. 
First do you want to use "rest" like 
`case 27: return {U"trace", {U"str", U"rest"}};` 
?
Secondly, how would the return true look like ?

**Comment by ahakanbaba** [comment]
After we resolve the last questions about how to write `case 27: return {U"trace", {U"str", U"obj"}};` I will write some documentation and this should be ready to merge.

File: core/desugarer.cpp (hunk 1 above)
Comment (sparkprime) [reply core/desugarer.cpp]: These are just giving the parameter names on the Jsonnet side, which matter because of being able to do std.md5(str="foo"), and otherwise aren't specified anywhere.  There is no equivalent for the return value, I just mentioned it because I was thinking of ways to avoid the 2nd param, which might get a bit annoying because of the need to put a ) some distance away, when you insert the std.trace.  However my suggestions are also a bit obtuse, since this really has nothing to do with assert or conditionals :)

Comment (sparkprime) [reply core/desugarer.cpp]: `case 27: return {U"trace", {U"str", U"rest"}};` would be fine

Comment (sbarzowski) [reply core/desugarer.cpp]: 👍 for `case 27: return {U"trace", {U"str", U"rest"}};`. Functions that provide alternative semantics can be later implemented as regular jsonnet functions (non-builtin).

**Comment by ahakanbaba** [comment]
I have just seen the go port
https://github.com/google/go-jsonnet
Who is expected to duplicate the new functionality there ?

**Comment by sbarzowski** [comment]
> I have just seen the go port
> https://github.com/google/go-jsonnet
> Who is expected to duplicate the new functionality there ?

It should be quite easy. Just a few lines added to this file: https://github.com/google/go-jsonnet/blob/master/builtins.go. The only non-obvious part is extracting the location - and for that `e.trace.loc` should work. If you could do it, that would be awesome and if not, I'll handle it, no worries.

File: doc/docs/stdlib.html [hunk 3]
```diff
@@ -618,3 +618,19 @@ <h4 id="mergePatch">std.mergePatch(target, patch)</h4>
 <p>Applies <code>patch</code> to <code>target</code> according to <a
 href="https://tools.ietf.org/html/rfc7396">RFC7396</a></p>
 
+<h3 id="Debugging">Debugging</h3>
+
+<h4 id="trace">std.trace(str, rest)</h4>
+
```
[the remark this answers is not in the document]
ahakanbaba: I was hoping you could cut a release after this merges. Can I call it 0.10.1 ?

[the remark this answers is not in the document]
sbarzowski: @sparkprime decides about these things. 

It would probably be 0.11 since new features were added. We would also need to add documentation for #481 before. I'm all for frequent releases, but the current procedure for new releases is a bit tedious AFAIK.

sbarzowski:
> About the go port, Is there a document describing the reasoning behind it ? 

A few reasons:
* The community around Kubernetes and Cloud generally (who are the main users of jsonnet currently) likes Go. I hope it will be more approachable this way and more people will contribute.
* It's easier to avoid security issues with Go. 
* We can use Golang GC instead of implementing our own.
* It was an opportunity to clean some things up and exhaustively test the implementation.
* I would say the Go version is quite a bit simpler and easier to experiment with.
* There are plans to build more tooling on top of it - and with Go it can be a bit easier. For now there is only an extremely primitive linter (a stub really) but it's a start. 

@sparkprime could probably come up with more things.

> Does it run any faster ? If the performance is better we would be interested in giving it a try.

It depends. There are things that are order of magnitude faster in go-jsonnet. Mostly because of more builtins and better desugaring of some code structures. But there are also some things that are slower.

> Also I recommend to redirect new features to go-jsonnet and stop accepting PRs to C++ except bugfixes from now on. Double committing can be really strenuous.

@sparkprime announced a plan to eventually deprecate C++ version, but some things need to happen first (fully shared tests, some remaining issues in go-jsonnet). Actually double committing is not that much of an issue - tests and stdlib are shared, and implementation details can differ. Only new language features need to be implemented in both places and adding such things is a pretty big deal anyway.

sparkprime:
The security concerns about people running Jsonnet on untrusted code on the server side was probably the biggest motivator for go-jsonnet.  It was started by jbeda of heptio but most of the work was done by sbarzowski.

I find go-jsonnet generally about 2x slower in my usage.  You could certainly experiment with it though, the commandline binary is a drop-in replacement.

It really needs someone to go over it with a profiler and tweak some of the code at a microscopic level, I suspect that would get a chunk of performance back.  The next thing would be to implement optimizations at the Jsonnet language level, and only implement them in go-jsonnet.

I would indeed like to cut over to add features to go-jsonnet only, but it's not quite ready yet.  In particular, the ASTs do not record comments and whitespace, which is bad for tooling.

I think we can merge this now, right sbarzowski?

sbarzowski:
Yes, LGTM.

sparkprime:
Thanks @ahakanbaba I think a lot of people will find this useful!

ahakanbaba:
@sparkprime  awesome. 
Could you please generate a new release so that we can also start using it ?

ahakanbaba:
@sparkprime Do you plan to generate a release anytime soon ? It would be great for and and make it easier for us to use std.trace().

abourget:
Any way to get timings of what's inside the `trace` (time to compute the `rest`) ?

sbarzowski:
@abourget No, but it's an interesting idea and probably quite easy to implement. (Though it may be tricky to interpret the results sometimes due to caching and laziness, at least this `std.traceWithTime` should probably always actually evaluate its second argument). You can create a separate issue so it doesn't get lost.  

Also if you are having performance issues feel free to report performance issues (especially if the problem is also happening with go-jsonnet).

PR #504, opened 2018-04-22
Add std.parseHex std.parseOctal via generalization of std.parseInt

sparkprime:


doc/ref/stdlib.html
```diff
@@ -713,6 +713,49 @@ <h4 id="parseInt">std.parseInt(str)</h4>
...
+      <h4 id="parseOctal">std.parseOctal(str)</h4>
+    </div>
+    <div style="clear: both"></div>
+  </div>
+</div>
+
+<div class="hgroup">
+  <div class="hgroup-inline">
+    <div class="panel">
+      <p>Parses an unsigned octal integer, which must begin with '0', from the input string.</p>
```
sbarzowski: Why to use `0` prefix at all in such a function? `std.parseOctal('755')` seems clear enough to me.
Same applies to parseHex.

Issue #499, opened 2018-04-18
RFE: parser and formatter hexfloat support


jwise:
For a project I'm currently working on, we need our JSON datasets to support bit-precise floating point numbers to manage some tests; we've extended our JSON parser to support C99/C++11-style hexfloats, and we'd like to extend Jsonnet the same way.  A hexfloat looks like: `0x1.921fb54p+1` (to represent pi).

I propose to make the following changes:

- Change the lexer and parser to support hexfloats.  It's possible that hexfloat parsing could be implemented as part of `std`, but it seems like this probably better lives just in the parser itself -- ergonomically, after all, these are just a different format of literals.
- No change is needed to the formatter's Unparser, since it spits out numbers as they came in from the input.
- Modify `jsonnet_unparse_number` to take a `bool hexfloat` parameter, which controls whether it is willing to output hexfloats.  If it is, and a number cannot be represented precisely in five or so digits of decimal, then it preferentially outputs a hexfloat.
- Modify the `vm` to plumb that `hexfloat` parameter through.
- And, finally, modify the `jsonnet eval` tool to take a `hexfloat` command line option.  We go through all this plumbing trouble, of course, because hexfloats are not valid real JSON -- but, some floats cannot be represented precisely in JSON, no matter how many digits we include.

I'm working with the NVIDIA legal folks to get a CLA signed on this so I can add these.  Does this sound like something that would be accepted upstream if I wrote a PR for it?

sparkprime:
I though that all floats could be represented in decimal if you have enough digits.  There's a link in the jsonnet_unparse_number implementation to a paper about it.

sparkprime:
(ignoring nan, inf, etc)

jwise:
I think this is, in theory, true, but it might not be possible with all libcs out there -- I suspect that you have to be very careful in how you implement a `strtod` / `strtof` to get guaranteed bit-accurate results that can produce every bit-combination of a float from decimal input.  (Note that the man page for `strtod`, on my Mac laptop, permits hexfloats, interestingly.)  So, even if Jsonnet is perfectly clear about what it is outputing, the libc on the other end might not be able to get the LSBs out.  (We are using Jsonnet to specify configurations for hardware, and so we want to be able to do bit-exact details of register values.)  So hexfloat out would be really useful for us to feed the variety of models that we have.

[...]

PR #517, opened 2018-05-17
Hexfloat support for parser; formatter

jwise:
This pull request has two independent commits to add support for hexfloats each to the output formatter and to the parser.  (This is an implementation of #499.)  There are tests available for the parser.  It fairly trivially passes hexfloat support back to the C++11 back end, which has native support.  

We anticipate heavily using the hexfloat-input support, whereas the hexfloat-output support would be "nice to have" for our use cases.


jwise:
Hmm, the compilers that Travis is using on Linux are _really_ old.  I'll fix the stderr issue in the morning ... and I'll also have a look at the CLA issue in the morning too (I have it all signed, but there might need to be further diagnostics).  Any chance we could get the Travis compiler version on Linux bumped to something more modern?

jwise:
@googlebot I signed it!

jwise:
Ping @sparkprime .  Would be interested to know if any parts of this would be accepted upstream -- we'd really like to avoid having to maintain an internal branch!

sparkprime:
Sorry for the delay.  So in C++ it's quite easy to parse a hexfloat, but what about Go?

sparkprime:
Basically I'm ok with extending the language to allow people to configure things very precisely (with hexfloats) as long as it doesn't make the language overly difficult to implement.

sparkprime:
As for the output, a global switch doesn't feel quite right.  We should look into the custom manifesting approach and see if it can be made to work.  An alternative is to use a native function to do the manifesting.  And also it seems as long as you're not losing precision with the decimal output, and if it's just for machine reading, maybe you don't need it at all.

sparkprime:
I found this: https://github.com/golang/go/issues/125

sparkprime:
Oh and the fact that we get hex integers for free with this is pretty cool

jwise:
Sorry for the delay.

I agree that "probably" we do not need hexfloat manifest support (though, perversely, the thing that seems the least useful is also the thing that is the least pain to implement, because it does not force itself on the golang implementation).

I did notice that Go does not have native hexfloat input support.  golang/go#12518 is better than last I checked, which was when I saw the "output" form of hexfloat, golang/go#8072, in which the Go team saw a user request and summarily gave the friendly response of "Not worth doing."

Thinking about it, it should not be terribly difficult to implement, especially limited to the "normalized" form (0x1.[...]p[...]); one could first (trivially) parse the mantissa, convert that to a double, and then add the exponent.  If one so desired to support denormals perfectly, it would be slightly more difficult, but not at all impossible.  So this should be very doable in Go, even in versions of Go that are not yet in the future.

sparkprime:
Closing this until we have a Golang solution. In all likelyhood we will deprecate the C++ solution before Go supports this though!

PR #519, opened 2018-05-18
[Tests] Go implementation shouldn't go over stack limit

sbarzowski:
Before it was verified that it goes over the limit...


sbarzowski:
@sparkprime 
Needed for https://github.com/google/go-jsonnet/pull/221

PR #525, opened 2018-06-10
Balance square-brackets in grammar spec

lihaoyi:
Looks like one closing bracket ran away from the `array` definition to go hang out with the `object index (general2)` definition

Also the comma separating the comprehensions in the `objinside` should be outside the `objlocal`s square brackets, since you don't need any objlocals in order to put a `,` between the `[x]: y` and the for comprehension expressions


lihaoyi:
I signed it

sparkprime:
Nice, thanks

PR #535, opened 2018-06-25
cmake fixes

asherikov:
This patch fixes compilation & testing using cmake and contains a few more minor corrections. The main issue with the current CMakeLists.txt is that `std.jsonnet.h` header is generated by an executable compiled in the same project, which is not supported by cmake, and the generated string does not contain null terminator.

---

- CMakeLists.txt
    - Respect CMAKE_BUILD_TYPE

- test_suite/CMakeLists.txt
    - fix path to jsonnet binary
    - workaround for nonportable tmp dir generation
    - add fmt_tests & fmt_idempotence_tests

- stdlib/CMakeLists.txt
    - generate std.jsonnet.h using cmake
    - add missing null terminator in std.jsonnet.h


sparkprime:
Hi, this looks useful. Can you sign the CLA please? I'd like to get this into the next release. thanks!

asherikov:
Sorry, it's too minor to bother.

PR #538, opened 2018-07-04
fix example oo-contrived.jsonnet.golden

EBazarov:
consider example old_g should have value equals to 102

### DIFF
--- a/examples/oo-contrived.jsonnet.golden
+++ b/examples/oo-contrived.jsonnet.golden
@@ -3,7 +3,7 @@
     "f": 5,
     "g": 105,
     "old_f": 2,
-    "old_g": 105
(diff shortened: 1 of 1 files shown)


EBazarov:
I signed it!

sbarzowski:
I'm pretty sure it was correct. The fields accessed through `super` still belong to the same object. Being able to override the fields used in parent objects is pretty much the point.

Technically, you can think of objects as having layers. `self` means grabbing a field from the outermost layer and `super` takes it from the layer under the one in which `super` is used. Of course if the field is not found, it subsequently searches the deeper layers.

sparkprime:
Agreed. These files are all automatically checked and currently pass. This PR causes the test to fail.

PR #539, opened 2018-07-05
Changed to use multi-stage docker build

gambrose:
Multi stage build to make final image smaller.

```
> docker images
REPOSITORY                       TAG                 IMAGE ID            CREATED             SIZE
jsonnet                          latest              20460ec97937        18 minutes ago      16.2MB
sparkprime/jsonnet               latest              097d0a351c9a        8 days ago          246MB
```

Seems to only need libstdc++ to run, but only done limited testing, so would be worth confirming before merging.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM alpine:latest
+FROM alpine:latest AS builder
 
 RUN apk -U add build-base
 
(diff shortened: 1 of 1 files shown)


gambrose:
I signed it!

sparkprime:
Thanks!

Issue #316, opened 2017-06-01
Safe way to detect whether a native function exists before trying to call it


anguslees:
Currently calling `std.native("noexist")` explodes with a runtime error.

A jsonnet library may want to try a native function if it exists, or fall back to some safe/slower alternative.   It would be neat if there was some way to do this.  (This is not hypothetical, we are about to have exactly this situation in ksonnet-lib)

I suggest adding a `std.hasNative("foo") -> bool` function, but another fine alternative would be to make `std.native("noexist")` just return `null` and leave it up to the jsonnet caller to check/assert the result.  (Trying to call a `null` function will be just as fatal (so no change there), but the error message might be less helpful without an explicit assert.)

Note a similar problem exists with things like external variables (`std.extVar("noexist")` explodes).  Returning `null` here seems like a good choice - so perhaps we should just do that for `std.native` too for consistency...  Any opinions before I go and propose a PR?

sparkprime:
The use case of falling back to a non-native version of a function is quite compelling.  I quite like the idea to return null as well.  I think all you have to do is find the error message in vm.cpp, and instead of throwing an exception, set scratch to makeNull() and return nullptr.

std.extVar is a different beast though because the set of variables is specified on the jsonnet commandline.  That means not only can a script can do different things depending on how it's invoked, but the fact that it does so can be buried deep in the code since extVar is effectively a global variable.  This is why there are no defaults to std.extVar -- the "not found" error forces you to acknowledge the existence of the variable at the top level.

An alternative to std.extVar is the top level arguments (TLAs), where since the variables have to be threaded through the code explicitly and are therefore visible at the top level.  Those can be given default values so that covers the use case you're talking about.

parano:
@anguslees @sparkprime any followups on this issue? I'm having a very similar use case here and would love to know if you have found a workaround or if you need help with the PR.

sparkprime:
I haven't got any bandwidth right now, scrambling to get stuff done before I go on vacation for 2 weeks at the end of the month.  However, it is an easy PR to make std.native return null which I think is good enough.  The null error will point to the specific place in the code where you called the null and so I don't think it will be hard to debug it.  Since std.native is an advanced feature, we can trust people to do an assert to encapsulate the error.

Don't touch std.extVar though.

parano:
@sparkprime thanks for the reply, I will submit a PR later today

sbarzowski:
Done in #529.

PR #550, opened 2018-08-17
std.extVar return null instead of raise an Excpetion

eloycoto:
At the moment the std.extVar raises an exception if the variable does
not exits, this change return null if the variable is not set, the same
behaviour as PR #529

This allows to users to set default variables from outside and the uers
that want to have a error can do a conditional condition.

Usage examples:
```
breo.acalustra.com ➜  jsonnet git:(master) ✗ ./jsonnet -V y=1 -e 'if std.extVar("x") != null then 1'
null
breo.acalustra.com ➜  jsonnet git:(master) ✗ ./jsonnet -V x=1 -e 'if std.extVar("x") != null then 1'
1
breo.acalustra.com ➜  jsonnet git:(master) ✗ ./jsonnet -V y=1 -e 'std.extVar("x")'
null
```

Signed-off-by: Eloy Coto <eloy.coto@gmail.com>
CLA is signed.


seh:
Since it's possible to supply null as a value, callers then can't distinguish between a value not being set and a value being set to null. Is that acceptable?
```
jsonnet --ext-code v=null --exec '{v: std.extVar("v")}'
{
   "v": null
}
```

sbarzowski:
Please see comments:
- https://github.com/google/jsonnet/issues/316#issuecomment-306069666
- https://github.com/google/jsonnet/issues/316#issuecomment-398604467

This feature would require more discussion.

eloycoto:
thanks @sbarzowski ;-)

sparkprime:
Sorry, it is this way deliberately :)

PR #553, opened 2018-09-20
std.parseJson

sbarzowski:


sbarzowski:
I still need to add tests and implement error handling. On seemed to work on a few tests I manually entered.

sbarzowski:
@sparkprime  This change is ready for review.
The json library I used unfortunately requires a newer gcc than what is provided by travis. I changed the travis configuration to use a newer compiler.  When I was at it, I configured travis gcc/osx to actually use gcc (`gcc` is an aliast for `clang` on OS X, it's necessary to use e.g. `gcc-4.9` to use the real gcc).

I also extracted `std.trace` tests from `stdlib.jsonnet`, because they change the output every time something is added there - so it's better to keep them isolated.

core/vm.cpp
```diff
@@ -1308,6 +1312,77 @@ class Interpreter {
...
+                filled = true;
+                break;
+
+            case json::value_t::null:
+                attach = makeNull();
+                filled = true;
+                break;
+
+            case json::value_t::array:{
+                attach = makeArray(std::vector<HeapThunk *>{});
```
sparkprime: you could initialize it to the right size here

core/vm.cpp
```diff
@@ -1308,6 +1312,77 @@ class Interpreter {
...
+                for (auto it = v.begin(); it != v.end(); ++it) {
+                    auto *thunk = makeHeap<HeapThunk>(idJsonObjVar, nullptr, 0, nullptr);
+                    obj->compValues[alloc->makeIdentifier(decode_utf8(it.key()))] = thunk;
+                    otherJsonToHeap(it.value(), thunk->filled, thunk->content);
+                }
+            } break;
+
+            case json::value_t::discarded: {
+                // ???
+                assert(false);
```
sparkprime: abort()

core/vm.cpp
```diff
@@ -1308,6 +1312,77 @@ class Interpreter {
...
+                        makeHeap<HeapThunk>(idArrayElement, nullptr, 0, nullptr));
+                    otherJsonToHeap(v[i], arr->elements[i]->filled, arr->elements[i]->content);
+                }
+            } break;
+
+            case json::value_t::object: {
+                attach = makeObject<HeapComprehensionObject>(
+                    BindingFrame{}, jsonObjVar, idJsonObjVar, BindingFrame{});
+                filled = true;
+                auto *obj = static_cast<HeapComprehensionObject *>(attach.v.h);
```
sparkprime: Can't you make a regular simple object here?

CMakeLists.txt
```diff
@@ -87,6 +87,7 @@ link_directories(${GLOBAL_OUTPUT_PATH})
 include_directories(
     include
     third_party/md5
+    third_party/
```
sparkprime: I think setup.py needs something too

core/vm.cpp
sbarzowski: I don't see any reason why not. I used `HeapComprehensionObject`, because it is used here: https://github.com/google/jsonnet/blob/v0.11.2/core/vm.cpp#L1347. Do you remember why `HeapComprehensionObject` was used there?

sparkprime: Oh it's because with HeapSimpleObject you have to make AST objects and those don't get garbage collected until the interpreter exits.  So using HeapComprehensionObject is basically a hack to minimize that (it's still leaky because it needs to make ids and those are not GC'd either).

sbarzowski: so we're leaving `HeapComprehensionObject`, I guess

sbarzowski:
@sparkprime I fixed the merge conflict. Anything else blocking this?

PR #562, opened 2018-10-10
Initial implementation of 'switch' syntax.

kinghajj:
The Jsonnet language is clearly inspired from functional programming. It currently lacks, however, something analogous to a C-style `switch` or ML/Rust `match`. So, I set out to implement one. This PR isn't ready to merge as-is (see the notes below,) but meant to open a discussion for whether the community views such a feature as worthwhile.

This is purely syntactic sugar. A construct like

	switch x for
	if 'hello' then 42
	if 'world' then 37
	else 54;

is transformed by the parser into an equivalent if-else chain

	if x == 'hello' then
	  42
	else if x == 'world' then
	  37
	else
	  54;

which is evaluated by the VM normally. For simplicity, an `else` clause is mandatory, and there's a limit of 1024 cases.

There are still some potential issues to resolve before merging:

- The branch for fixing indentation is currently empty. The formatting tests pass, so I'm not sure if one is needed.
- The compiler pass visitor just explores the internal if chain, but I'd guess that it should visit the logical structure instead.
- The code for constructing the if-chain should probably move into the desugarer module.
- As written, the desugared if-chain simply re-uses the pivot AST. So if it is an expression instead of a variable lookup, it could be expensive to re-compute it for each case.

### DIFF
--- a/core/ast.h
+++ b/core/ast.h
@@ -39,6 +39,7 @@ enum ASTType {
     AST_BINARY,
     AST_BUILTIN_FUNCTION,
     AST_CONDITIONAL,
+    AST_SWITCH,
     AST_DESUGARED_OBJECT,
     AST_DOLLAR,
     AST_ERROR,
@@ -74,6 +75,7 @@ static inline std::string ASTTypeToString(ASTType type)
         case AST_BINARY: return "AST_BINARY";
         case AST_BUILTIN_FUNCTION: return "AST_BUILTIN_FUNCTION";
         case AST_CONDITIONAL: return "AST_CONDITIONAL";
+        case AST_SWITCH: return "AST_SWITCH";
         case AST_DESUGARED_OBJECT: return "AST_DESUGARED_OBJECT";
         case AST_DOLLAR: return "AST_DOLLAR";
         case AST_ERROR: return "AST_ERROR";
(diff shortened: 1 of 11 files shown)


kinghajj:
I have signed the CLA, just didn't have my Github handle listed.

seh:
In your example, does that mean that we can't match `x` against the string "else," it being a keyword, or does the "else" there instead designate an expected value for `x`? The original example used 54 as the value when `x` doesn't match, so it's probably the former. Would there be an alternative clause to handle the case where `x` doesn't match any of the cases, like this?
```
switch x for {
  hello: 42,
  world: 37,
} else 54,
```

sparkprime:
In regular objects it's not allowed because else is a keyword:

```
$ jsonnet -e '{ else: 5 }'
STATIC ERROR: <cmdline>:1:3-7: unexpected: else while parsing field definition
```

So basically in this case, else is the default case and if you actually want the string else you would put it in quotes (just like with an object).

kinghajj:
The problem I see with limiting cases to string literals, is that it prevents developing libraries that export variants as fields.

```
// stuff.libsonnet
{
    mode: {
        party: 'stuff/mode/party',
        study: 'stuff/mode/study',
    }
}

// main.jsonnet
local stuff = import "stuff.libsonnet";

switch something for
if stuff.mode.party then 'example'
if stuff.mode.study then 'another'
else 'nothing';
```

This way, the particular value of the `party`/`study` fields can change, but the interface remain constant. That wouldn't be possible if the cases had to be string literals only.

sparkprime:
You could do expressions in there, it'd be like object syntax: `[stuff.mode.party]: foo`

sparkprime:
@sbarzowski remember discussing this?

sbarzowski:
No, I don't remember talking about this before.

I think the @sparkprime's solution can be implemented without extending the language. Something like:
```
switch x for {
  hello: 42,
  world: 37,
}
```
is equivalent to:
```
{
  hello: 42,
  world: 37,
}[x]
```

If we want the default value we would need a function to handle that, it would be something like:
```
std.switch(x, {
  hello: 42,
  world: 37,
}, 54)
```

and `std.switch` would be 
```
function(x, options, default) if options.HasField(x) then options[x] else default
```

Perhaps it's worth it to use a function also when there is no default value, to more clearly express intent.

If it's a really common pattern, a special syntax may also be worth it. The @seh's proposal of `switch <expr> for <expr> else <expr>` makes sense to me (it may be slightly weird if the switch is itself inside an if, since we can get ambiguity issues like we experienced in #567).

I hope that one day a richer `match` syntax will come. But we should at least get deconstruction working for locals before we get into that.

Issue #557, opened 2018-10-03
std.escapeStringJson escapes ~ as if it was unicode


lihaoyi-databricks:
```bash
$ cat foo.jsonnet
std.escapeStringJson("~")

$ jsonnet foo.jsonnet
"\"\\u007e\""
```

I'd expect `~` to be left untouched, since it plays no role in any JSON syntax and doesn't affect quoting or escaping at all

sparkprime:
I believe it is intended to find all control characters. 159 is 9f here: https://en.wikipedia.org/wiki/Latin-1_Supplement_(Unicode_block)

126 should definitely not be included in that range, since it is printable.

PR #564, opened 2018-10-18
Change escape range to 0080-009F to match the latin-1 charset

seryl:
Fixed the escape range for printable json values and added associated tests.

Should address #557

### DIFF
--- a/stdlib/std.jsonnet
+++ b/stdlib/std.jsonnet
@@ -853,7 +853,7 @@ limitations under the License.
         '\\t'
       else
         local cp = std.codepoint(ch);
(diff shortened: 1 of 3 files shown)


stdlib/std.jsonnet
```diff
@@ -853,7 +853,7 @@ limitations under the License.
         '\\t'
       else
         local cp = std.codepoint(ch);
-        if cp < 32 || (cp >= 126 && cp <= 159) then
+        if cp < 32 || (cp >= 128 && cp <= 159) then
```
sparkprime: 127 is also a control character :)

Issue #549, opened 2018-08-02
std.parseInt('') and std.parseInt('-') should produce a clear error message


sbarzowski:
Current behavior:
```
$ ./jsonnet -e "std.parseInt('-')"
-0
```

```
$ ./jsonnet -e "std.parseInt('')"
RUNTIME ERROR: string bounds error: 0 not within [0, 0)
        std.jsonnet:84:8-14     thunk <a>
        std.jsonnet:84:8-21     function <anonymous>
        std.jsonnet:84:8-21     function <anonymous>
        <cmdline>:1:1-17
```

sbarzowski:
Fixed in #569.

PR #569, opened 2018-10-21
Update std.parseInt error messages

sneivandt:
This change address issue #549 by adding more informative error messages when attempting to parse `''` or `'-'` as an integer.


sneivandt:
I signed it!

sparkprime:
Thanks, I'm going to merge this which will break tests, and then rebuild golden to fix it.

sparkprime:
Looks like it didn't break tests after all

Issue #567, opened 2018-10-19
Conditional bug in jsonnet.org/learning/tutorial.html


stephen-soltesz:
The conditionals.jsonnet example from https://jsonnet.org/learning/tutorial.html appears to include a mistake.

The given input jsonnet is:
```
local Mojito(virgin=false, large=false) = {
  // A local next to fields ends with ','.
  local factor = if large then 2 else 1,
  // The ingredients are split into 3 arrays,
  // the middle one is either length 1 or 0.
  ingredients: [
    {
      kind: 'Mint',
      action: 'muddle',
      qty: 6 * factor,
      unit: 'leaves',
    },
  ] + if virgin then [] else [
    { kind: 'Banks', qty: 1.5 * factor },
  ] + [
    { kind: 'Lime', qty: 0.5 * factor },
    { kind: 'Simple Syrup', qty: 0.5 * factor },
    { kind: 'Soda', qty: 3 * factor },
  ],
  // Returns null if not large.
  garnish: if large then 'Lime wedge',
  served: 'Over crushed ice',
};

{
  Mojito: Mojito(),
  'Virgin Mojito': Mojito(virgin=true),
  'Large Mojito': Mojito(large=true),
}
```

And, the generated output for the "Virgin Mojito" does not include all virgin ingredients. Instead it only contains (full output abbreviated):
```
{
  ... snip ...
  "Virgin Mojito": {
    "garnish": null,
    "ingredients": [
      {
        "action": "muddle",
        "kind": "Mint",
        "qty": 6,
        "unit": "leaves"
      }
    ],
    "served": "Over crushed ice"
  }
}
```

This appears to be due to the precedence of the condition and `+` operators.
```
 ] + if virgin then [] else [
    { kind: 'Banks', qty: 1.5 * factor },
  ] + [
    { kind: 'Lime', qty: 0.5 * factor },
    { kind: 'Simple Syrup', qty: 0.5 * factor },
    { kind: 'Soda', qty: 3 * factor },
  ],
```

Placing the conditional block in parentheses generates the expected output, but since I'm still learning, I'm not sure if there's a preferable way.
```
 ] + (if virgin then [] else [
    { kind: 'Banks', qty: 1.5 * factor },
  ]) + [
    { kind: 'Lime', qty: 0.5 * factor },
    { kind: 'Simple Syrup', qty: 0.5 * factor },
    { kind: 'Soda', qty: 3 * factor },
  ],
```

Is this a bug? What's the best fix?

[...]

PR #571, opened 2018-10-23
Fix conditionals example.

stephen-soltesz:
This change updates the conditionals example to exclude only the alcoholic ingredient when `virgin=true`.

Before this change, all ingredients after the `else` condition were also excluded when `virgin=true`

Fixes issue: https://github.com/google/jsonnet/issues/567


sbarzowski:
Thanks!

LGTM @sparkprime

sparkprime:
Yep, thanks!

PR #576, opened 2018-11-20
Refactor jsonnet desugaring to avoid stack overflow.

dmuir:
Default thread stacks on Windows are 1Mb and on Macos are 512Kb
(except for the main thread which is 8Mb) and we've been seeing stack
overflows on these platforms.  It appears that recursively calling
desugar consumes over 17Kb of stack, so the recursion doesn't have to go
very deep before it overflows the stack.  This refactoring shrinks the
stack usage and lets our jsonnet run on Mac & Windows.

### DIFF
--- a/core/desugarer.cpp
+++ b/core/desugarer.cpp
@@ -393,6 +393,277 @@ class Desugarer {
         return super_vars;
     }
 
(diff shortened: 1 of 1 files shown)


sparkprime:
LGTM, thanks

PR #577, opened 2018-11-22
Add std.encodeUTF8, std.decodeUTF8

sbarzowski:


core/vm.cpp
```diff
@@ -1291,6 +1293,51 @@ class Interpreter {
...
+    const AST *builtinDecodeUTF8(const LocationRange &loc, const std::vector<Value> &args)
+    {
+        validateBuiltinArgs(loc, "decodeUTF8", args, {Value::ARRAY});
+
+        std::string byteString;
+
+        HeapArray *arr = static_cast<HeapArray *>(scratch.v.h);
+        for (auto *thunk : arr->elements) {
+            if (!thunk->filled) {
+                stack.newCall(loc, thunk, thunk->self, thunk->offset, thunk->upValues);
```
sbarzowski: @sparkprime I need your help here. It doesn't work this way, and I'm not sure what is the right way. I need to get the (evaluated) elements of this array.

sparkprime: I think you're going to need to record intermediate state on the stack while you recurse into the thunk elements. This is not easy.  Look at how std.filter is implemented natively -- it has a special stack frame it pushes.  If this puts you off implementing this natively in C++ I would understand it :)

sparkprime: dmuir just did something similar for std.join in another PR

sbarzowski:
@sparkprime I changed it to use frames and it seems to work. The @dmuir's work was indeed very helpful.

core/vm.cpp
```diff
@@ -1308,6 +1314,61 @@ class Interpreter {
...
+    const AST *decodeUTF8(void)
+    {
+        Frame &f = stack.top();
+        const auto& elements = static_cast<HeapArray*>(f.val.v.h)->elements;
+        while (f.elementId < elements.size()) {
+            auto *th = elements[f.elementId];
+            if (th->filled) {
+                auto b = th->content;
+                if (b.t != Value::NUMBER) {
+                    throw makeError(stack.top().location, "Decoded byte was not a number");
```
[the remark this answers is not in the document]
sbarzowski: Done

I've also added tests for these errors

PR #580, opened 2018-12-03
Implement several stdlib functions with native builtins.

dmuir:
Migrate several function from std.jsonnet to native implementations.  This improves performance of our jsonnet conversion by about 30%.

### DIFF
--- a/core/vm.cpp
+++ b/core/vm.cpp
@@ -858,6 +858,12 @@ class Interpreter {
         builtins["native"] = &Interpreter::builtinNative;
         builtins["md5"] = &Interpreter::builtinMd5;
         builtins["trace"] = &Interpreter::builtinTrace;
+        builtins["splitLimit"] = &Interpreter::builtinSplitLimit;
+        builtins["substr"] = &Interpreter::builtinSubstr;
+        builtins["range"] = &Interpreter::builtinRange;
+        builtins["strReplace"] = &Interpreter::builtinStrReplace;
+        builtins["asciiLower"] = &Interpreter::builtinAsciiLower;
+        builtins["asciiUpper"] = &Interpreter::builtinAsciiUpper;
     }
 
     /** Clean up the heap, stack, stash, and builtin function ASTs. */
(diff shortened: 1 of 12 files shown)


core/vm.cpp
```diff
@@ -1308,6 +1314,121 @@ class Interpreter {
...
+        return nullptr;
+    }
+
+    const AST *builtinAsciiLower(const LocationRange &loc, const std::vector<Value> &args)
+    {
+        validateBuiltinArgs(loc, "asciiLower", args, {Value::STRING});
+        const auto *str = static_cast<const HeapString *>(args[0].v.h);
+        UString new_str(str->value);
+        for (int i = 0; i < new_str.size(); ++i) {
+            if (new_str[i] > 64 && new_str[i] < 91) {
```
sbarzowski: Nit: Could you use 'a' and 'z' instead of numbers?

sbarzowski:
Thanks a lot! @dmuir btw, since you mentioned that performance became an issue for you, have you tried the Go implementation (github.com/google/go-jsonnet)? It's quite a bit faster (partially due to native implementations of builtins).

LGTM @sparkprime.

seh:
Do we have any benchmarks that would demonstrate the improvement here? I'm not doubting these native functions are faster; I'd just like to see how much faster they are.

stdlib/std.jsonnet
```diff
@@ -35,18 +35,6 @@ limitations under the License.
   toString(a)::
     if std.type(a) == 'string' then a else '' + a,
 
-  substr(str, from, len)::
```
sparkprime: If you leave these in, it won't break the Go implementation.  They should be overridden by the C++ ones in desugarer.cpp

sparkprime:
I'm ok with this but in the long term it'd be good to grease the wheels with moving people over to the Go implementation.  I suppose that native C binding is the next step?  That would enable moving the python module over as well.

stdlib/std.jsonnet
sparkprime: (All you have to do is revert the changes in this file.)

core/vm.cpp
dmuir: I've made the suggested change and updated my fork.  If there's something else I need to do to update this pull request, please let me know (I'm a github newbie).

stdlib/std.jsonnet
sparkprime: Ok the problem there is that this code (in desugarer.cpp) is adding them even though they're already in std_obj->fields.

```
        // Bind 'std' builtins that are implemented natively.
        DesugaredObject::Fields &fields = std_obj->fields;
        for (unsigned long c = 0; c <= max_builtin; ++c) {
            const auto &decl = jsonnet_builtin_decl(c);
            Identifiers params;
            for (const auto &p : decl.params)
                params.push_back(id(p));
            fields.emplace_back(ObjectField::HIDDEN,
                                str(decl.name),
                                make<BuiltinFunction>(E, encode_utf8(decl.name), params));
        } 
        fields.emplace_back(
            ObjectField::HIDDEN, str(U"thisFile"), str(decode_utf8(ast->location.file)));
```

What's necessary is to build a DesugaredObject::Fields with only the fields we want.

Something like this (I haven't tried compiling it):

```
        std::unordered_map<Identifier, DesugaredObject::Field> fields;
        for (const auto &jsonnet_field : std_obj->fields) {
            fields[jsonned_field.name] = jsonnet_field;
        }
        for (unsigned long c = 0; c <= max_builtin; ++c) {
            const auto &decl = jsonnet_builtin_decl(c);
            Identifiers params;
            for (const auto &p : decl.params)
                params.push_back(id(p));
            fields[decl.name] = DesugaredObject::Field(
                ObjectField::HIDDEN, str(decl.name),
                make<BuiltinFunction>(E, encode_utf8(decl.name), params));
        } 
        fields[decl.name] = DesugaredObject::Field(
            ObjectField::HIDDEN, str(U"thisFile"), str(decode_utf8(ast->location.file)));
        std_obj->fields.clear();
        for (const auto &item : fields) {
            std_obj->fields.push_back(item.second);
        }
```

dmuir: I've implemented something along these lines (though the details are different).  PTAL.

PR #584, opened 2018-12-13
Support for jpathdir in python bindings

parijatgarg:
The command line version allows the specification of a jpathdir, which lets one use jsonnet libraries. The same support was missing in the python bindings. Adding that.


parijatgarg:
I signed it!

sparkprime:
Thanks for sorting out the CLA stuff

PR #597, opened 2019-01-23
Mixin example

schmidh:
The ingredients lists should be flat and not lists of lists.


schmidh:
I signed it!

schmidh:
Made email address associated with my Github account public.

sparkprime:
Thanks

PR #599, opened 2019-01-29
Emphasise set requirements

Code0x58:


doc/ref/stdlib.html
```diff
@@ -1625,8 +1630,8 @@ <h4 id="setUnion">std.setUnion(a, b)</h4>
   <div class="hgroup-inline">
     <div class="panel">
       <p>
-        Set union operation (values in either a or b).  Note that + on sets will simply concatenate
-        the arrays, possibly forming an array that is not a set (due to not bein ordered without
+        Set union operation (values in either a and/or b).  Note that + on sets will simply concatenate
```
sbarzowski: Small nit. "either and/or" looks a bit weird to me. "Either or" seems to mean XOR (https://english.stackexchange.com/questions/13889/does-either-a-or-b-preclude-both-a-and-b). But perhaps it can be written simply as "values appearing in  <code>a</code> or in <code>b</code>".  

Adding an example could help too (see std.join for formatting).

doc/ref/stdlib.html
```diff
@@ -1568,6 +1568,11 @@ <h3 id="sets">Sets</h3>
   <div class="hgroup-inline">
     <div class="panel">
       <p>Sets are represented as ordered arrays without duplicates.</p>
+      <p>
+        Note that the <code>std.set*</code> functions rely on the uniqueness and ordering
+        on arrays passed to them to work. This can be guarenteed by using <code>std.set(arr)<code>.
+        If that is not the case, the functions will quielty return non-meaningful results.
```
sparkprime: s/quielty/quietly

PR #600, opened 2019-01-31
Better wording/formatting in the docs

sbarzowski:


doc/ref/stdlib.html
```diff
@@ -1704,7 +1704,7 @@ <h4 id="base64">std.base64(v)</h4>
   <div class="hgroup-inline">
     <div class="panel">
       <p>
-        Encodes the given value into a base64 string.  The encoding sequence is A-Za-z0-9+/ with =
+        Encodes the given value into a base64 string.  The encoding sequence is <pre>A-Za-z0-9+/</pre> with <pre>=</pre>
```
Code0x58: would `<code>` be better here than `<pre>`? It looked like `<pre>` wasn't inline

Issue #583, opened 2018-12-07
RFE: (std.sortFn) sort function that accepts a function, result of function is what is evaluated for sorting


ghostsquad:
It would be nice to be able to sort based on an arbitrary function. Example use case:
```
{
  foos: [
    { foo: 3 },
    { foo: 1 },
    { foo: 2 },
  ],

  sorted: std.sortFn(function(f) f.foo, $.foos),
}
```
expected result:
```
sorted: [
  { foo: 1 },
  { foo: 2 },
  { foo: 3 },
]
```

seh:
You're in luck: It's already there, as [the second parameter to `std.sort`](https://github.com/google/jsonnet/blob/master/stdlib/std.jsonnet#L1093).

This came up recently [in the "ksonnet" channel](https://kubernetes.slack.com/archives/C6JLE4L9X/p1543368494217000?thread_ts=1543364503.216000&cid=C6JLE4L9X) of the "Kubernetes" Slack team.

sbarzowski:
It's not mentioned in the docs (https://jsonnet.org/ref/stdlib.html#sort)), though. We should add it.

ghostsquad:
Ah, ya, the docs don't mention it. So I had no idea!

sbarzowski:
It's documented for some time now (thanks to @redbaron). I'm going to close this issue.

PR #602, opened 2019-02-11
Document keyF arguments for std.set* and std.sort functions

redbaron:
Follow up for https://github.com/google/jsonnet/pull/481

xref: https://github.com/google/jsonnet/issues/583


sparkprime:
Thanks

Issue #601, opened 2019-02-11
setUnion should use merge sort


redbaron:
right now setUnion builds new `std.set()` which  is is quite expensive comparing to merge sorting already sorted lists

PR #603, opened 2019-02-11
Merge sort in setUnion

redbaron:
Fixes https://github.com/google/jsonnet/issues/601

### DIFF
--- a/stdlib/std.jsonnet
+++ b/stdlib/std.jsonnet
@@ -1162,8 +1162,22 @@ limitations under the License.
     std.length(std.setInter([x], arr, keyF)) > 0,
 
   setUnion(a, b, keyF=id)::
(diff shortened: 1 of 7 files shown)


sparkprime:
You're missing a comma at the end which is breaking the tests.  Also you're gonna break a load of the "error" tests because of the line numbers changing in the tracebacks.  But you can regenerate those golden files easily with ./refresh_golden.sh which can take a list of jsonnet files

redbaron:
Thanks, fixed (hopefully).

It is also somewhat breaking change: std.setUnion() worked with arbitrary arrays before, but now it expects `set.setUnion(std.set(), std.set())`

sparkprime:
Yeah I think it's ok as it doesn't need to be bug compatible.  I'll document it in the release notes.

PR #611, opened 2019-02-26
CMake: Add USE_SYSTEM_GTEST option (default to OFF).

cdluminate:
On Debian/Ubuntu based systems downloading googletest every time is unnecessary. With the patch one could do as follows and avoid downloading it:

```shell
cmake . -GNinja -DUSE_SYSTEM_GTEST=ON
```


sparkprime:
Looks good, thanks

PR #616, opened 2019-03-06
Remove LIB_JSONNET_VERSION from libjsonnet.h

sbarzowski:
It could be very confusing if header is linked with older/newer
version of the library. We have jsonnet_version function which
returns the version of the actual library underneath.

So unless it's some standard thing to provide the version of the
header separate from the version of the actual library, I think it's
just a bad idea to include it, especially considering go-jsonnet C bindings.

### DIFF
--- a/include/libjsonnet.h
+++ b/include/libjsonnet.h
@@ -24,8 +24,6 @@ limitations under the License.
  * of using the library.
  */
 
-#define LIB_JSONNET_VERSION "v0.12.1"
-
 /** Return the version string of the Jsonnet interpreter.  Conforms to semantic versioning
  * http://semver.org/ If this does not match LIB_JSONNET_VERSION then there is a mismatch between
  * header and compiled library.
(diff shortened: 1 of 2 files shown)


include/libjsonnet.h
```diff
@@ -24,8 +24,6 @@ limitations under the License.
  * of using the library.
  */
 
-#define LIB_JSONNET_VERSION "v0.12.1"
-
 /** Return the version string of the Jsonnet interpreter.  Conforms to semantic versioning
  * http://semver.org/ If this does not match LIB_JSONNET_VERSION then there is a mismatch between
```
sparkprime: The idea was that you would be able to compare the two to check the .so was correct, but that may not be particularly useful, because we basically have to be backwards compatible anyway.  I don't mind removing it, but this comment needs updating. Also setup.py expects to find the version in this specific file.

sparkprime:
Closing this because there are even more places pulling this out of the header now

Issue #618, opened 2019-03-07
setDiff causes 'max stack frames exceeded' error


asobti:
Jsonnet version

```
$ jsonnet --version
Jsonnet commandline interpreter v0.11.2
```

Minimal repro case:

```
local set1 = std.set(std.range(1000, 1500));
local set2 = std.set();

{
  result: std.setDiff(set1, set2)
}
```

Causes error:

```
$ jsonnet setdiff.jsonnet -o test.json
RUNTIME ERROR: max stack frames exceeded.
	std.jsonnet:1092:15-25	thunk <l>
	std.jsonnet:1097:34	
	std.jsonnet:1097:20-64	thunk <rest>
	std.jsonnet:1098:60-64	
	std.jsonnet:1098:20-65	thunk <left>
	std.jsonnet:1100:16-20	thunk <arr>
	std.jsonnet:1093:19-22	
	std.jsonnet:1093:8-23	thunk <a>
	std.jsonnet:1093:8-28	function <anonymous>
	std.jsonnet:1093:8-28	function <anonymous>
	...
	std.jsonnet:763:5-28	function <anonymous>
	std.jsonnet:1110:5-26	function <anonymous>
	std.jsonnet:1113:5-40	function <anonymous>
	setdiff.jsonnet:1:14-44	thunk <set1>
	setdiff.jsonnet:5:23-27	thunk <a>
	std.jsonnet:1149:9	thunk <a>
	std.jsonnet:1149:5-24	function <aux>
	std.jsonnet:1149:5-24	function <anonymous>
	setdiff.jsonnet:5:11-34	object <anonymous>
	During manifestation	
```

sbarzowski:
Thanks for the report. I was able to reproduce it on current master. I think I know what's going on.

sparkprime:
This had nothing to do with setDiff, it was actually set() that wasn't working if the input was already sorted.

sparkprime:
E.g. this works fine:

```
local set1 = std.range(1000, 1500);
local set2 = [];

{
  result: std.setDiff(set1, set2)
}
```

But Stan fixed it anyway by changing std.sort to use merge sort instead of quicksort for large input arrays.

PR #619, opened 2019-03-07
Change sort algorithm

sbarzowski:
Current std.sort implementation recurses may recurse N
times for the array of size N. For example for sorted or
reverse sorted data. This causes problems with std.set for example.

This change implements mergesort which always divides the array
in half, so the depth of recursion is always reasonable.

For smaller arrays quicksort is still used (it is faster).


sbarzowski:
@sparkprime 

Fixes #618

sparkprime:
Lol nice, I did not expect this to be the problem.

PR #622, opened 2019-03-18
Update virgin mojito

sparkprime:


doc/_includes/examples/mixins.jsonnet
```diff
@@ -20,10 +20,10 @@ local NoAlcohol = {
     ] else [
       ingredient,
     ],
-  ingredients: [
+  ingredients: std.flattenArrays([
```
sbarzowski: Alternatively `std.flatMap(Substitute, super.ingredients)` could be used I think. But I'm not sure if it would be more clear for newcomers.

sbarzowski:
LGTM

PR #626, opened 2019-03-27
CMake: add {,SO}VERSION properties for libjsonnet.so

cdluminate:
Tracking ABI/API changes with SOVERSION is an important part for distribution packaging.


sparkprime:
thanks

PR #627, opened 2019-03-28
CMake: various improvements

cdluminate:
Changes:
1. Use `GNUInstallDirs` instead of hardcoded installation path
2. Also install the static library
3. Remove `install(DIRECTORY include DESTINATION include)` because it installs garbage to the dest dir. Headers will be installed along with the `libjsonnet` target through `PUBLIC_HEADER` property.


sparkprime:
Thanks!

PR #628, opened 2019-03-29
Make ... optional in YAML output

### LINKED ISSUE #614: Add --- on top of each YAML object in .yml files
Hello,

Would it be possible to prepend each YAML object in `.yml` files with `---`? I personally need to read several YAML files from a `cat` command so without this it doesn't work as I don't know where each YAML object starts.

Thanks
Geoffrey

sparkprime:
Fixes #614

### DIFF
--- a/stdlib/std.jsonnet
+++ b/stdlib/std.jsonnet
@@ -1004,13 +1004,13 @@ limitations under the License.
           std.join('\n' + cindent, lines);
     aux(value, [], ''),
 
-  manifestYamlStream(value, indent_array_in_object=false)::
+  manifestYamlStream(value, indent_array_in_object=false, c_document_end=true)::
     if std.type(value) != 'array' then
       error 'manifestYamlStream only takes arrays, got ' + std.type(value)
     else
       '---\n' + std.join(
         '\n---\n', [std.manifestYamlDoc(e, indent_array_in_object) for e in value]
-      ) + '\n...\n',
+      ) + if c_document_end then '\n...\n' else '\n',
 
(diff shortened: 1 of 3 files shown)


stdlib/std.jsonnet
```diff
@@ -1004,13 +1004,13 @@ limitations under the License.
           std.join('\n' + cindent, lines);
     aux(value, [], ''),
 
-  manifestYamlStream(value, indent_array_in_object=false)::
+  manifestYamlStream(value, indent_array_in_object=false, disable_c_document_end=false)::
```
sbarzowski: It is not really important, but perhaps instead of `disable_c_document_end=false`, `c_document_end=true` or `with_c_document_end=true` would be more natural (right now it's a bit like double negation).

sparkprime: I think the reason I did it that way is that I'm used to proto3 messages where the default is always false.  Any idea if the python style guide says anything about preferring falsey defaults?

sbarzowski: I don't think Python style guide says anything about that. Python standard library uses both True and False as default values. 

In case of proto3 I guess it mostly has to do with interoperability between versions and with protobuf not sending these default values over the wire (which is a good reason to have one fixed default value per type or else the sender's default may be changed to the receiver's default).  This is a concern for over-the-wire protocols, but not really for a library function.

sparkprime: Done.

sbarzowski:
LGTM

PR #634, opened 2019-04-15
[Website] Fix emscripten settings, so demo doesn't crash in Firefox

### LINKED ISSUE #621: libjsonnet.js: function Pointer_stringify was removed / UTF8ToString is not exported
Hi

I built the javascript library by running ```make libjsonnet.js```. When I tried to use the resulting lib like shown in https://jsonnet.org/js/demo.js, I got an error stating that the function Pointer_stringify was removed and that UTF8ToString should be used instead. However this function is not exported.

I think the solution would be adding the function to the Makefile in line 134?
JS_RUNTIME_METHODS = 'EXTRA_EXPORTED_RUNTIME_METHODS=["cwrap", "_free", "getValue", "lengthBytesUTF8", "_malloc", "Pointer_stringify", "setValue", "stringToUTF8", "addFunction", "UTF8ToString"]'

Should I open a pull request?

Thanks

### LINKED ISSUE #633: Live examples don't work on firefox
Hi! I just discovered jsonnet but I stumbled on some issues when I tried to play with the examples on the site (https://jsonnet.org).  If I take the first example and start adding `person3: self.person1 { name: "Cat" },` it seems to work find at first but after a few keystroke, I get the output:
```
abort("Cannot enlarge memory arrays. Either (1) compile with  -s TOTAL_MEMORY=X  with X higher than the current value 16777216, (2) compile with  -s ALLOW_MEMORY_GROWTH=1  which allows increasing the size at runtime but prevents some optimizations, (3) set Module.TOTAL_MEMORY to a higher value before the program runs, or (4) if you want malloc to return NULL (0) instead of this abort, compile with  -s ABORTING_MALLOC=0 "). Build with -s ASSERTIONS=1 for more info.
```
image

And then it doesn't work anymore, even if I restore the original example.

Some observations:
- I'm not sure but this looks like an error message from emscripten
- if I just paste the new line instead of typing it in, it works fine

I'm using Firefox 62 on Fedora but I also tested with firefox 66 and got the same issue.  I does seem to work fine on Chromium.

sbarzowski:
Fixes #633
Also fixes #621, because I wasn't able to run it otherwise.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -32,8 +32,8 @@ OPT ?= -O3
 CXXFLAGS ?= -g $(OPT) -Wall -Wextra -Woverloaded-virtual -pedantic -std=c++0x -fPIC -Iinclude -Ithird_party/md5 -Ithird_party/json
 CFLAGS ?= -g $(OPT) -Wall -Wextra -pedantic -std=c99 -fPIC -Iinclude
 MAKEDEPENDFLAGS ?= -Iinclude -Ithird_party/md5 -Ithird_party/json
-EMCXXFLAGS = $(CXXFLAGS) -g0 -Os --memory-init-file 0 -s DISABLE_EXCEPTION_CATCHING=0 -s OUTLINING_LIMIT=10000 -s RESERVED_FUNCTION_POINTERS=20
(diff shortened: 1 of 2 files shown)


sparkprime:
LGTM (on vacation right now)

PR #635, opened 2019-04-15
Add std.reverse

sbarzowski:


stdlib/std.jsonnet
```diff
@@ -1132,6 +1132,10 @@ limitations under the License.
     local bytes = std.base64DecodeBytes(str);
     std.join('', std.map(function(b) std.char(b), bytes)),
 
+  reverse(arr)::
+    local l = std.length(arr);
+    std.makeArray(l, function(i) arr[l - i - 1]),
```
seh: I haven't benchmarked this, but what do you think about this condition?
```
if l < 2 then arr
else
  local bias = l - 1;
  std.makeArray(l, function(i) arr[bias - i]),
```

sbarzowski: Well, I guess it's going to be slightly faster, but the real cost here is function call and thunk creation. A builtin is really a way to go if performance of this function is critical.

PR #636, opened 2019-04-16
Move kubecfg to bitnami org

mkmik:
Kubecfg now lives under the bitnami github org.
GitHub redirects old links but I guess it's a good idea to fix the links anyway (also to break the ties with the deprecated ksonnet tool)


sbarzowski:
Thanks!

sbarzowski:
(The changes will be visible when the website gets regenerated).

PR #643, opened 2019-04-30
Fix std.substr out of bounds inconsistency

sbarzowski:
Make both Jsonnet and C++-native implementation clip
the substring which go beyond the end of string.


sbarzowski:
It will fix https://github.com/google/go-jsonnet/issues/272 after the submodule is updated. BTW go-jsonnet should also have a builtin implemention of this function, but first things first.

PR #645, opened 2019-05-01
Fix #644: make string interpolation error msg clearer.

### LINKED ISSUE #644: String interpolation message unclear when values inconsistent
Interpolating the following string in jsonnet `'%(var) rate > 50%' % some_variable,` would produce a stacktrace with ` RUNTIME ERROR: Not enough values to format, got 1` 
Which is valid but not really clear on what the problem is.
Same for string like: `'%(var) rate > %(var)' % some_variable,`

Could we make the error message easier to understand and hint more about the problem so it's easier to resolve in large codebases?

Maybe something like `String interpolation failed. Expected x values but received y` ?

uberspot:


sbarzowski:
@sparkprime Could you take a look?

stdlib/std.jsonnet
```diff
@@ -670,7 +670,7 @@ limitations under the License.
           local tmp2 = if code.prec == '*' then {
             j: tmp.j + 1,
             prec: if tmp.j >= std.length(arr) then
-              error 'Not enough values to format: ' + std.length(arr)
+              error ('Not enough values to format: ' + std.length(arr) + ', expected ' + j)
```
sparkprime: should be `tmp.j` not `j` here

stdlib/std.jsonnet
```diff
@@ -682,7 +682,7 @@ limitations under the License.
             if j2 < std.length(arr) then
               arr[j2]
             else
-              error 'Not enough values to format, got ' + std.length(arr);
+              error ('Not enough values to format: ' + std.length(arr) + ', expected ' + j);
```
sparkprime: should be `j2`, not `j`

stdlib/std.jsonnet
```diff
@@ -660,7 +660,7 @@ limitations under the License.
           local tmp = if code.fw == '*' then {
             j: j + 1,
             fw: if j >= std.length(arr) then
-              error 'Not enough values to format: ' + std.length(arr)
+              error ('Not enough values to format: ' + std.length(arr) + ', expected ' + j)
```
sparkprime: we don't actually know how many are expected because we haven't gotten to the end of `codes` yet.  So it could say `"expected at least " + j` or something like that (the same in the other cases).

sparkprime: Alternatively could use something like `std.length(std.filter(codes, function(e)std.isObject(e)))`

sparkprime: Oh you also have to count the ones with `*` twice, so that doesn't work.

sparkprime: I guess could map it to either 0 1 or 2 and then use std.foldl with a +

uberspot: `expected at least` still makes sense and is helpful so I changed it to that for now and fixed the other variable issues as well. :) Thanks for reviewing.

sparkprime:
Thanks

PR #654, opened 2019-05-15
Fix %e handling of 0

### LINKED ISSUE #640: '%e' % 0 gives error
To reproduce:
```
$ ./jsonnet -e "'%e' % 0"
RUNTIME ERROR: overflow
	std.jsonnet:546:34-55
	std.jsonnet:546:24-70	thunk <exponent>
	std.jsonnet:549:42-50
	std.jsonnet:549:30-51	thunk <mantissa>
	std.jsonnet:551:24-32	thunk <n__>
	std.jsonnet:533:39-42	thunk <n>
	std.jsonnet:791:17
	std.jsonnet:791:8-19	thunk <a>
	std.jsonnet:791:8-31	function <anonymous>
	std.jsonnet:791:8-31	function <anonymous>
	...
	std.jsonnet:467:12	function <aux>
	std.jsonnet:471:7-17	function <padding>
	std.jsonnet:475:7-38	function <pad_left>
	std.jsonnet:695:15-39	thunk <s_padded>
	std.jsonnet:701:55-63	thunk <v>
	std.jsonnet:701:11-64	function <format_codes_arr>
	std.jsonnet:701:11-64	function <format_codes_arr>
	std.jsonnet:749:7-48	function <anonymous>
	std.jsonnet:227:7-23	function <anonymous>
	<cmdline>:1:1-9
```

Taking the log of 0 at https://github.com/google/jsonnet/blob/c323f5ce5b8aa663585d23dc0fb94d4b166c6f16/stdlib/std.jsonnet#L546
seems to be the cause.

sparkprime:
Fix #640

### DIFF
--- a/stdlib/std.jsonnet
+++ b/stdlib/std.jsonnet
@@ -546,7 +546,12 @@ limitations under the License.
       local exponent = if n__ == 0 then 0 else std.floor(std.log(std.abs(n__)) / std.log(10));
       local suff = (if caps then 'E' else 'e')
                    + render_int(exponent, 3, 0, false, true, 10, '');
-      local mantissa = n__ / std.pow(10, exponent);
+      local mantissa = if exponent == -324 then
+        // Avoid a rounding error where std.pow(10, -324) is 0
+        // -324 is the smallest exponent possible.
+        n__ * 10 / std.pow(10, exponent + 1)
+      else
+        n__ / std.pow(10, exponent);
       local zp2 = zero_pad - std.length(suff);
       render_float_dec(mantissa, zp2, blank, sign, ensure_pt, trailing, prec) + suff;
 
--- a/stdlib/std.jsonnet
+++ b/stdlib/std.jsonnet
@@ -543,7 +543,7 @@ limitations under the License.
 
     // Render floating point in scientific form
     local render_float_sci(n__, zero_pad, blank, sign, ensure_pt, trailing, caps, prec) =
-      local exponent = std.floor(std.log(std.abs(n__)) / std.log(10));
+      local exponent = if n__ == 0 then 0 else std.floor(std.log(std.abs(n__)) / std.log(10));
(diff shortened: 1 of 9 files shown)


sparkprime:
@sbarzowski it's better than nothing, but you're probably right about needing a builtin at least for this little bit of functionality.

sbarzowski:
We already have a builtin it's called `std.exponent` and using it there fixes the 0 case. I haven't figured out yet the mantissa part though (the only issue is that it's all base-2 and we need to be really careful with conversion not to get out of range).

sparkprime:
They're both base 2, I was wondering whether that's a deal breaker.

```
dcunnin@dcunnin1:~$ jsonnet -e '[std.mantissa(100), std.exponent(100)]'
[
   0.78125,
   7
]
```

sbarzowski:
Well, in case of exponent it's a matter of multiplying by a constant factor of `std.log(2)/std.log(10)`. So for example: `std.floor(std.exponent(100)*std.log(2)/std.log(10))`.  Using that may be enough to solve the problem just for `0`.

For converting mantissa we need to divide it by `2**base2exponent/10**base10exponent`. It's only the floor in the expression above which moves it away from 1, so it should be in [1, 10) range. IIRC I still had issues when I just computed that with pow.

sparkprime:
`std.floor(std.exponent(99)*std.log(2)/std.log(10))` is 2 though when it should be 1. The problem is it's already taken the ceiling before we convert the base.

sparkprime:
@sbarzowski I'm willing to declare victory on this now.

sbarzowski:
Yeah, this is the case I mentioned here:
> So after multiplication new "mantissa" is in range [0.5, 10). Exponent may need to be adjusted [incremented] so that it [the mantissa] falls in [0.1, 1).

Anyway I think it's good enough for now.

PR #656, opened 2019-05-16
Separate jsonnet fmt into its own executable.

sparkprime:
@sbarzowski 

Note that editor integrations will have to change "jsonnet fmt" to "jsonnetfmt" but this does mean you won't have to worry about the go executable.

Some matching changes are needed on the go side (remove eval, adjust --help, etc).


sbarzowski:
Nice!

PR #665, opened 2019-06-02
Add RegEx support using RE2

dcoles:
Introduces 5 new built-in methods to the stdlib:
- `regexFullMatch(pattern, str)` -- Full match regex
- `regexPartialMatch(pattern, str)` -- Partial match regex
- `regexQuoteMeta(str)` -- Escape regex metachararacters
- `regexReplace(str, pattern, to)` -- Replace single occurance using regex
- `regexGlobalReplace(str, pattern, to)` -- Replace globally using regex

Since both `regexFullMatch` and `regexPartialMatch` can perform captures
these functions return a "match" object upon match or `null` otherwise.
For example:

```
$ ./jsonnet -e 'std.regexFullMatch("h(?P<mid>.*)o", "hello")'
{
   "captures": [
      "ell"
   ],
   "namedCaptures": {
      "mid": "ell"
   },
   "string": "hello"
}
```

Introduces a dependency on [RE2 2019-06-01](https://github.com/google/re2/releases/tag/2019-06-01).
Builds tested using make, CMake and Bazel on Ubuntu 18.04.


dcoles:
Hey @googlebot, I signed it. 😄

dcoles:
Linux and OSX build fails on `make` due to a missing dependency on RE2. For Linux I can update `.travis.yml` to pull in `libre2-dev`, but I'm not sure the right approach for OSX.

Bazel and CMake builds should be unaffected since they pull in the RE2 source via Git.

sparkprime:
It looks like on OSX you can do:

```
addons:
  homebrew:
    packages:
    - re2
```

Judging by https://docs.travis-ci.com/user/installing-dependencies/#installing-packages-on-macos and http://macappstore.org/re2/

rbauduin:
I could really use this, is it still planned to include this?

nojvek:
Will there be any progress on landing regex support for jsonnet? Almost been an year. Just wondering what the status of this PR is.

sbarzowski:
We're happy to go forward if the PR author (or someone else) is willing to pick this up again and fix the build issues.

nojvek:
Thanks for quick response @sbarzowski.

I see there are some merge conflicts and travis is failing with `The command "sudo -E apt-get -yq --no-install-suggests --no-install-recommends $(travis_apt_get_options) install g++-4.9 libre2-dev" failed and exited with 100 during .` I'll wait for a couple of days if @dcoles is going to pick it up again, otherwise I'll try to pick this up and give it a shot next week.

sparkprime:
I assume that anyone picking this up would have to find a compatible solution on the Go side too? (I.e. the same regex syntax and semantics)

sbarzowski:
Golang stdlib is "almost" re2. 

My stance on this is that regexps are too useful to miss out on them completely, but fully portable, formally-defined solutions are not available, so I'd be willing to compromise and have best-effort compatibility.

sparkprime:
That may be necessary if there is no alternative.  We would need to allow the jsonnet code to know what implementation it's running as so it can choose a regex for that implementation.  I also wonder what the impact would be for the unofficial implementations in rust, scala, etc.

sbarzowski:
I think, in practice what people almost always want are very simple regexps which look the same everywhere, so probably checking platform is not going to be necessary for what the people actually want...

Duologic:
I rebased this PR here: https://github.com/google/jsonnet/pull/1039

MrG9090 (approved):

MrG9090 (approved):

VILJkid:
Eagerly waiting for this patch to get released...

PR #668, opened 2019-06-04
Makefile: Fix missing SONAME on shared libraries

dcoles:
Most distributions expect that a SONAME for ABI versioning[1].
Set the SONAME to the same values set by CMake build.

1. https://dev.gentoo.org/~zmedico/portage/doc/ch07s04.html

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -40,6 +40,9 @@ LDFLAGS ?=
 
 SHARED_LDFLAGS ?= -shared
 
+VERSION = 0.13.0
+SOVERSION = 0
+
 ################################################################################
 # End of user-servicable parts
 ################################################################################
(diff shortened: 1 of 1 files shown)


dcoles:
My Mac-foo isn't all that strong, so I'm not quite sure that `-install_name` on OSX is exactly the analog of `-soname` on other Unix-es.

sparkprime:
I just fixed the CMakeLists.txt version as that is an obvious problem (not sure if it's enough of a problem to cut a new release for).

dcoles:
Thanks! Dropped the CMakeLists.txt commit.

Honestly, since it only affects the installed library path (and no change to SOVERSION), it's probably not going to cause anyone much grief. Nice to fix, but no need to worry about making a new release immediately.

PR #675, opened 2019-06-22
Add cpp/ to cmake and libs and install targets to Makefile

brettviren:
Changes for

* `cmake` to build targets under `cpp/` (`libjsonnet++.so` and tests)
* `make` to build libs in addition to bins by default
* `make install` target to install bins, libs and incs

FYI, I signed a CLA.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -95,7 +95,8 @@ include_directories(
     include
     third_party/md5
     third_party/json
-    core)
+    core
+    cpp)
 
 if (BUILD_TESTS)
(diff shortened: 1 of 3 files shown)


sparkprime:
Thanks, I dunno what's going on with the travis tests.  It doesn't look related.

brettviren:
Can someone merge this?

Also related: https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=931011

sparkprime:
Thanks

PR #692, opened 2019-08-06
Properly detect a system GTest

refi64:
I'm hoping to try making a Fedora COPR package for jsonnet, but right now using a system GTest requires the given hardcoded path. I switched it to use CMake's find_package, which AFAIK should detect GTest on pretty much every distro.


sbarzowski:
Seems more reasonable than hardcoding the path. I believe this is supposed to be "the right way" from CMake perspective.

There was a test failure which looked like Travis being Travis, so I have just restarted it.

sparkprime:
Thanks

PR #697, opened 2019-08-13
provide a multiline function example

jacobstr:
Prompted by the slack discussion here:
https://kubernetes.slack.com/archives/C6JLE4L9X/p1565650799070500

Sometimes you want a function that returns a 'scalar' value, but
benefits from some intermediate calculations. In my case specifically,
I was trying to build a list of availability zones in GKE from a struct
describing accelerator-types and their gpu zones in us-central1.

```
// Mapping of accelerator types to availability zones in GCP us-central1.
local global_pools = {
  v100: {
    gpu_zones: ['a', 'b', 'c', 'f'],
  },
  p100: {
    gpu_zones: ['c', 'f'],
  },
  // This will result in an assertion failure.
  // t4: {
  //   gpu_zones: ['a', 'b'],
  // },
};

// When building a regional GKE cluster that uses GPUs, one needs to
// specify the available zones in accordance with intersection of all
// desired accelerator-type-friendly zones at cluster creation. This is
// because the cluster's zones and the zones of any node pool must
// coincide.
local gpu_zones(pools) = (
  local zones = [pools[name].gpu_zones for name in std.objectFields(pools)];
  local available_zones = std.foldl(std.setInter, zones, zones[0]);
  assert std.length(available_zones) >= 2 : 'Too few zones support chosen accelerators';
  available_zones
);

gpu_zones(global_pools)
```

### DIFF
--- a/examples/functions.jsonnet
+++ b/examples/functions.jsonnet
@@ -18,6 +18,13 @@ limitations under the License.
 // Default arguments are like Python:
 local my_function(x, y=10) = x + y;
 
+// Define a local multiline function.
+local multiline_function(x) =
+  // One can nest locals.
+  local temp = x * 2;
+  // Every local ends with a semi-colon.
+  [temp, temp + 1];
+
 local object = {
   // A method
   my_method(x): x * x,
@@ -27,6 +34,8 @@ local object = {
   // Functions are first class citizens.
   call_inline_function:
     (function(x) x * x)(5),
+
+  call_multiline_function: multiline_function(4),
 
   // Using the variable fetches the function,
   // the parens call the function.
--- a/examples/functions.jsonnet.golden
+++ b/examples/functions.jsonnet.golden
@@ -2,6 +2,7 @@
   "call": 12,
   "call_inline_function": 25,
   "call_method1": 9,
+  "call_multiline_function": [8, 9],
   "len": [
     5,
     3


examples/functions.jsonnet
```diff
@@ -18,6 +18,13 @@ limitations under the License.
 // Default arguments are like Python:
 local my_function(x, y=10) = x + y;
 
+// Define a local multiline function. Parenthesis are optionally used as
+// a block delimiter - e.g. curly braces in C-like languages.
+local multiline_function(x) = (
```
sparkprime: You don't need the parens and I think as long as there is indentation, there's not much point adding them.  You can just put the ; immediately after the temp.

jacobstr: Happy to remove these. I imagine the parser isn't whitespace sensitive and knows to terminate the `multiline_function` definition when it encounters the terminal `temp`. What is the most precise word that encompasses the terminal statements here: e.g.  it could be either `temp` or `std.sort(...)` or `{"foo": "bar"}`.

sparkprime: It's the `;` that ends it, so maybe:
(I'm keeping this very short so it looks good on the website)

```
// Define a local multi-line function.
local multiline_function(x) =
  local temp = x * 2;  // One can nest locals.
  temp;  // Every local ends with a semicolon.
```

sparkprime: It'd make more sense if you actually did something with temp, even if it's just `[temp, temp + 1]` or something like that.

sparkprime: You can paste it into the tutorial to make sure it actually fits

jacobstr: Updated according to these suggestions.

sparkprime:
Awsome, thanks!  Can you do the CLA?  I can't merge without that.

sparkprime:
(See the GoogleBot message further up the thread)

jacobstr:
Yep - working through that with the legal department. Cheers!

jacobstr:
@googlebot I signed it!

PR #703, opened 2019-08-23
Add tar of executables to github releases.

PaulRudin:
This gets travis to add 2 tar files containing the two executables to github releases; one for linux and one for mac. I've arbitrarily picked the gcc build binaries, since it seems there's little point in including the ones built with both compilers. 

This should happen anytime a new tag is pushed.

In order to for this to work someone with appropriate permissions needs to create a github personal access token with (at least) the public_repo permission in the github UI, and then add this as a secure environment variable called JSONNET_GITHUB_TOKEN in the project's settings in the travis dashboard.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,5 @@
 language: generic
+
 matrix:
   include:
     - os: linux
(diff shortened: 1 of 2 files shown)


PaulRudin:
@googlebot I signed it!

sparkprime:
I set it up in travis, or at least I think I did.

PR #704, opened 2019-08-26
Ensure 'jsonnetfmt -i' does not modify timestamps if target file unchanged

dcwangmit01:


rustycl0ck:
LGTM

This will help a lot in getting rid of the workarounds to handle this in `Makefile`s.
Thanks

sbarzowski:
Thanks! It's very nice that you also added a test for that.

PR #718, opened 2019-10-10
fix dangling pointer warning

BenTheElder:
fixes:
```
external/jsonnet/cmd/jsonnet.cpp:298:31: warning: object backing the pointer will be destroyed at the end of the full-expression [-Wdangling-gsl]
            const char *arg = next_arg(i, args).c_str();
                              ^~~~~~~~~~~~~~~~~
```
(I think, admittedly untested)
seen when compiling with bazel for https://github.com/kubernetes/test-infra


sparkprime:
Thanks

BenTheElder:
thanks for the quick response!

Issue #722, opened 2019-10-29
Outputs invalid JSON int values (erroneously uses locale formatting)


davidljung:
When I pass in a Jsonnet string like:
`{ type: "int", limited: true, min: 0, max: 2000, params: [ "RS34" ] }`

and execute it using the C++ API via evaluateSnippet(), the resulting string is:

`
      {
         "limited": true,
         "max": 2,000,
         "min": 0,
         "params": [
            "RS34"
         ],
         "type": "int"
      }
`

which isn't valid JSON (notice the extra comma for the thousands position for the 2000).

This is likely the result of using a std::ostream for formatting integers, without correctly setting the numpunct to empty (resulting in the comma for the current locale being used).  This is US English locale - I suspect for a Euro locale it might result in "2.000"!

sbarzowski:
Thank you very much for reporting this. This is exactly the kind of dependence of the environment that we are trying to avoid with Jsonnet, so I would like to fix it asap. Unfortunately, I wasn't able to reproduce the problem.

Could you provide us with the following information:
* Your OS
* Jsonnet version that you're using
* How did you install Jsonnet? Have you compiled it, used a precompiled version or perhaps fetched it from some public repository?
* The exact locale that you're using (output of `locale` command)

On my end I tried setting the following envvars:
```
export LC_ALL=en_US.UTF-8  
export LANG=en_US.UTF-8
```
And I ran your example, i.e.:
```
jsonnet -e '{ type: "int", limited: true, min: 0, max: 2000, params: [ "RS34" ] }'
```
However, the output didn't exhibit the problem you describe. I did that on OS X 10.13.6 with current `master` and older `v0.11.2` version of Jsonnet.

davidljung:
OS: Ubuntu 18.04 x86 64
Version: v0.14.0 771845d11919d62895a8cb478fb73fa1cd2d2bc1 (just cloned master today from GitHub)
Installation: I actually just ran make from the repo root, then linked directly into my app (didn't install on system)

I'm using the en_US.UTF-8 locale, but I don't think that is directly relevant.
My understanding of C++ locale management isn't 100%, but I believe that the default global C++ locale (as far as std::locate code is concerned) defaults to some kind of classic locale ("C"? "POSIX"? unsure).  This 'global' (from the process perspective) locale is used as a prototype when things like iostreams are created.

However, our app explicitly sets the process global locale according to user selected language & region (it is a multilingual app) using code similar to:

```c
[...]

PR #724, opened 2019-10-30
Set the stream locale to "C" locally

sbarzowski:
So that even if there is a global locale
set by the library user, Jsonnet output
is still the same.


sbarzowski:
Fixes https://github.com/google/jsonnet/issues/722.

It makes sense to add a regression test and check that we don't have the same problem in other places. Perhaps we can have a helper function to create stringstreams and clear the locale?

sparkprime:
The test failed on mac gcc.  It looked like a legitimate failure with the locale stuff but I didn't understand the error.  I ran it again anyway, just in case.

sparkprime:
Same thing:
```
terminate called after throwing an instance of 'std::runtime_error'
  what():  locale::facet::_S_create_c_locale name not valid
```

sparkprime:
My guess would be that it needs some packaged installed to get the right locale files. However I'm not sure why it works on clang but not gcc?  Maybe they use different standard libraries and the clang library bundles it.

sbarzowski:
I tried getting to the bottom of this again, but I'm still very confused. One idea that I haven't checked yet is that the locale is just not there, but clang-compiled-version falls back to something "gracefully" and gcc-compiled-version crashes.

X33x356 (approved):

johnbartholomew:
This still seems worth fixing. `std::locale::classic` / C locale should always be available I think. We just need the test itself to not rely on a particular locale that might not be available on the system where the test is running.

I've rebased to master and made some adjustments to the test, see https://github.com/google/jsonnet/compare/master...johnbartholomew:jsonnet:sbarzowski-fix-locale-dependence

johnbartholomew:
Re: the test failure on Mac (https://github.com/google/jsonnet/pull/724#issuecomment-564268750), it might be this? https://stackoverflow.com/a/58385001

Anyway, I think it can/should be avoided by just defining a custom locale facet for the test.

johnbartholomew:
As this PR has been open for a long time (4+ years), I'll just go ahead and rebase and merge this (with a few adjustments/additions)

PR #740, opened 2019-12-17
Makefile: Add DESTDIR

ndowens:
Add $(DESTDIR) alongside of $(PREFIX) as many packaging tools set DESTDIR to a chroot location

Signed-off-by: Nathan Owens <ndowens04@gmail.com>

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -126,12 +126,12 @@ endif
 default: jsonnet jsonnetfmt
 
 install: bins libs
-	mkdir -p $(PREFIX)/bin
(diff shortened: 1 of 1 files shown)


ndowens:
@googlebot I signed it!

ndowens:
By tools, I mean like Arch Linux PKGBUILD, VoidLinux's xbps-src. Debian's
APT.

DESTDIR gets defined as the path of CHROOT during packaging, so the files
doesn't install, until a user installs the created package.

sbarzowski:
Thanks!

PR #741, opened 2019-12-17
Redact stdlib line numbers in tests

sbarzowski:
Before this change, every time stdlib was modified
a bunch of tests would need to be updated, because
some line numbers in stack traces changed.

It may not be the most elegant way to do it, but
it is simple and I grew tired of fixing the tests
every time - also the noise makes it harder to spot
genuine changes to the Jsonnet output.


sbarzowski:
The test on the Mac was not passing previously, because Mac's `sed` does not expant "\t" into a tab character.

Issue #762, opened 2020-01-21
Resolve confused function usage 


seongahjo:
Hello, I'm happy to find the library that increase my productivity.
I had look up std, I found that some predicate functions is wrapped as "isxxx", just like isString.
But in some places like "substr", "map" use direct "type" function to check types.
I think it is better to use "isxxx" function.
I'd like to make pull request if you regard it as a problem.

sbarzowski:
Yes, that would indeed be better and PR like this would be much appreciated. `std.type(v) == 'xxx'` is an old pattern and `isXxx` is the new and better way.  

BTW such change may eventually lead to some performance gains if we add builtin implementations for `isXxx` functions (but don't worry about this part now).

PR #763, opened 2020-01-22
refactor type checking to use defined function (#762)

seongahjo:
All test passed

All type-checking are implemented by 'isxxx' function, excepts for "format_codes" functions.
They make errors when I change them.


sbarzowski:
This a very good change, thank you!

PR #771, opened 2020-02-22
cmake: Use CMAKE_CURRENT_SOURCE_DIR instead of CMAKE_SOURCE_DIR

hanazuki:
When Jsonnet CMake project is included in another project using `add_subdirectory`, `CMAKE_SOURCE_DIR` points to the root directory of the main project but not the root of the Jsonnet tree.
To avoid this problem, instead use `CMAKE_CURRENT_SOURCE_DIR`, which points to the directory where the current CMakeList resides.

### DIFF
--- a/core/CMakeLists.txt
+++ b/core/CMakeLists.txt
@@ -32,7 +32,7 @@ add_library(libjsonnet SHARED ${LIBJSONNET_HEADERS} ${LIBJSONNET_SOURCE})
 add_dependencies(libjsonnet md5 stdlib)
 target_link_libraries(libjsonnet md5 nlohmann_json::nlohmann_json)
 
-file(STRINGS ${CMAKE_SOURCE_DIR}/include/libjsonnet.h JSONNET_VERSION_DEF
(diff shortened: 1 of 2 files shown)


sbarzowski:
Makes sense, thanks.

The only problem is that the tests are failing in CI for this change. Do you know why? Can you fix it?

hanazuki:
I'm looking into the strange failure, but I don't know why it happens and even not yet able to reproduce it on my side. I'll keep investigating it.

hanazuki:
CI's still failing with this patch reverted.

This change in gtest is suspicious: https://github.com/google/googletest/commit/23b2a3b1cf803999fb38175f6e9e038a4495c8a5 
CMake and GCC in Travis look too old for the change.

sbarzowski:
The CI issue should be fixed now (https://github.com/google/jsonnet/pull/773). Sorry for the trouble and thanks for the tip about googletest!

Please rebase this PR when you have a moment.

hanazuki:
Thank you!

PR #785, opened 2020-03-17
cmake: Specify INTERFACE include directories for libraries

hanazuki:
Specify INTERFACE include directories for the libraries so that the libraries' dependent targets can use the proper include directory automatically.

This is useful when Jsonnet is used as a subproject in a CMake tree. A target that links to libjsonnet will be configured to use the proper include directory without explicit `target_include_directories`. This has an advantage that the consumer need not to know the internal directory layout of the Jsonnet tree.
```cmake
add_executable(my-project ...)
add_subdirectory(path/to/jsonnet EXCLUDE_FROM_ALL)
target_link_libraries(my-project libjsonnet)
# target_include_directories(my-project PRIVATE path/to/jsonnet/include)
```

### DIFF
--- a/core/CMakeLists.txt
+++ b/core/CMakeLists.txt
@@ -48,6 +48,8 @@ install(TARGETS libjsonnet
 	LIBRARY DESTINATION "${CMAKE_INSTALL_LIBDIR}"
 	ARCHIVE DESTINATION "${CMAKE_INSTALL_LIBDIR}"
 	PUBLIC_HEADER DESTINATION "${CMAKE_INSTALL_INCLUDEDIR}")
+target_include_directories(libjsonnet INTERFACE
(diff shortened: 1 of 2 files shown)


sbarzowski:
@hanazuki I'm not really a CMake expert, could you provide a release notes comment explaining how exactly it affects users (i.e. what they need/should/can change in their configuration)?

hanazuki:
Thank you for accepting my patch.

I suggest the release note entry would be:

The CMake library targets (`libjsonnet` and `libjsonnet++`) now publish the include directory required to find their interface headers (`libjsonnet.h` and `libjsonnet++.h`, resp.), so that it is not necessary to add target_include_directories to the consumers of the targets to compile with libjsonnet.

PR #793, opened 2020-04-03
Support for multiple jpathdirs in python bindings

ilpianista:
By using the command line tool it's possible to specify additional library search dir while the python bindings only support one additional search dir.

I made the `jpathdir` attribute a list, but I also preserved backward compatibility with previous versions.


ilpianista:
@googlebot I signed it!

sparkprime:
Looks good and I'm surprised it wasn't this way in the first place, actually.  Must have been an oversight on my part.

PR #794, opened 2020-04-04
throw error on shift with negative exponent

mikedanese:
Jsonnet spec:

> Operators <<, >>, &, ^, | and ~ first convert their operands to signed
> 64 bit integers, then perform the operations in a standard way, then
> convert back to IEEE double precision floating point.

ISO 9899:2011 6.5.7 Bit-wise shift operators:

> The integer promotions are performed on each of the operands. The type
> of the result is that of the promoted left operand. If the value of
> the right operand is negative or is greater than or equal to the width
> of the promoted left operand, the behavior is undefined.

Jsonnet spec says jsonnet performs operations in "a standard way" which
is not terribly specific. Alternatively, we could support negative
shifts. Right now, my locally compiled build just returns 0 for any
negative shfit.


sparkprime:
The C++ implementation just does a C++ shift with the given value as an int64, which is UB if it's negative.  What it actually seems to do is take the unsigned int mod 64 and shift by that, so -1 is equivalent to 63, etc.  The go implementation also lets the -1 underflow when it casts it, but `<< (2^64 - 1)` is not the same as `<< 63` in Go.  It just treats it as a very large shift so Go always gives 0.  I'm not sure what the JVM version does - quite possibly something different again.

```
dcunnin@dcunnin:~$ go-jsonnet -e '1 << -1'
0
dcunnin@dcunnin:~$ jsonnet -e '1 << -1'
-9223372036854775808
```

Javascript seems to behave like our C++ implementations except it wraps it at 32 instead of 64. Python throws an error: `ValueError: negative shift count` and Go panics if you give it a negative int (which go-jsonnet never does because it casts it).

So anyway, yes.  We need to decide what we want it to do, write it up in the spec and consolidate the implementations.  My vote is for raising an error because I don't see the usefulness of assigning any other meaning to negative shifts.

sbarzowski:
Error on negative shift sounds good to me. 

(Negative shifts could be reasonably interpreted as shift in the other direction. But I think it's safer to just make it an error).

mikedanese:
Error for negative shift sounds good. Are we also ok with standardizing the mod 64 normalization of rhs? This appears to be what's happening with shift counts greater than 63. 
 
```
$ cat bitshift.jsonnet 
local exps = [0, 1, 63, 64, 65];
local lhs = 1;
['%d >> %d = %d' % [lhs, rhs, lhs >> rhs] for rhs in exps] +
['%d << %d = %d' % [lhs, rhs, lhs << rhs] for rhs in exps]
$ jsonnet bitshift.jsonnet 
[
   "1 >> 0 = 1",
   "1 >> 1 = 0",
   "1 >> 63 = 0",
   "1 >> 64 = 1",
   "1 >> 65 = 0",
   "1 << 0 = 1",
   "1 << 1 = 2",
   "1 << 63 = -9223372036854776028",
   "1 << 64 = 1",
   "1 << 65 = 2"
]
```

The `1 << 63` result is interesting as well. It appears to be depending on two's complement representation of int64_t.

sparkprime:
Thanks Mike

PR #797, opened 2020-04-19
Adjust spec for recent relaxation of rules governing function parameters

sparkprime:
The behaviour is defined by desugaring into a form where all params have default expressions, and the default default is an expression that raises an error.  This actually simplifies the operational semantics rule.

### DIFF
--- a/doc/ref/spec.html
+++ b/doc/ref/spec.html
@@ -52,7 +52,7 @@ div.rules {
     \newcommand{\binary}[3]{ #1\texttt{ }#2\texttt{ }#3}
     \newcommand{\error}[1]{\texttt{error }#1}
     \newcommand{\false}{\texttt{false}}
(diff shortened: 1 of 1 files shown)


doc/ref/spec.html
```diff
@@ -879,6 +882,9 @@ <h3 id="core_syntax">Core Syntax</h3>
           The following binary operators are removed: <code>!=</code> <code>==</code> <code>%</code>
           <code>in</code>
         </li>
+        <li>All function parameters have default expressionss. If no default expression was given
```
sbarzowski, line 885: It took me a while to understand this item. What does it mean to "use an error to force the caller to specify the value"?

Maybe:
"All function parameters have default expressions. If no default expression was given in the original source, a dummy one, which evaluates to an error, is used. "

(Also a typo "expresionss".)

sparkprime: Ok so I can't string a sentence together.  But at least the semantics rules are correct? :)

Issue #799, opened 2020-04-20
Floating-point formatting does not show the negative sign


yangle:
```
$ echo '{x: -0.5, y: "%f" % self.x}' | jsonnet -
{
   "x": -0.5,
   "y": "0.500000"
}
$ jsonnet --version
Jsonnet commandline interpreter v0.15.0
```

sbarzowski:
Thanks a lot for reporting! Wow, I don't know how such a basic bug survived all these years. 

This happens only when the number is between in (-1, 0) range.
Here is the reason: https://github.com/google/jsonnet/blob/master/stdlib/std.jsonnet#L538. We expect that `sign(x) * floor(abs(x))` has the same sign as `x`. That's not true for numbers in (-1, 0) where `floor(abs(x)) = 0`.

yangle:
Great. Glad to help!

sbarzowski:
Fixed in #801

PR #801, opened 2020-04-23
Fix floating point formatting, #799

sparkprime:


sparkprime:
Took me a while to realise that render_int would actually take a floating point number and internally do the right thing with the sign

stdlib/std.jsonnet
```diff
@@ -535,7 +535,7 @@ limitations under the License.
       local whole = std.floor(n_);
       local dot_size = if prec == 0 && !ensure_pt then 0 else 1;
       local zp = zero_pad - prec - dot_size;
-      local str = render_int(std.sign(n__) * whole, zp, 0, blank, sign, 10, '');
+      local str = render_int(n__, zp, 0, blank, sign, 10, '');
```
sbarzowski, line 538: Could you modify the comment for render_int explaining what it does with non-whole numbers? This behavior is counter-intuitive, so it's better to have it spelled out explicitly.

test_suite/format.jsonnet
```diff
@@ -189,6 +189,7 @@ std.assertEqual(std.format('%-12.4f', [910.3]), '910.3000    ') &&
 std.assertEqual(std.format('%#.0f', [910.3]), '910.') &&
 std.assertEqual(std.format('%#.0f', [910]), '910.') &&
 std.assertEqual(std.format('%.3f', [1000000001]), '1000000001.000') &&
+std.assertEqual(std.format('%f', [-0.1]), '-0.100000') &&
```
sbarzowski, line 192: I would also add `std.format('%d', [-0.1])'. The result will be '-0', I think. Not sure if that's ok.

sparkprime: I think it's not OK because it's different to what Python does.

stdlib/std.jsonnet
sparkprime: Have attempted to do it less hackily now

stdlib/std.jsonnet
```diff
@@ -483,23 +483,36 @@ limitations under the License.
...
-      local aux(n) =
-        if n == 0 then
-          zero_prefix
+    // Render a sign & magnitude integer (radix ranges from decimal to binary).
+    // neg should be a boolean, and when true indicates that we should render a negative number.
+    // mag must always be a whole number >= 0, it's the magnitude of the integer to render
+    // min_chars must be a whole number >= 0
+    //   It is the field width, i.e. std.length() of the result should be >= min_chars
+    // min_digits must be a whole number >= 0. It's the number of zeroes to pad with.
+    // blank must be a boolean, if true a postive number has a ' ' in place of the '-'.
```
sbarzowski, line 492: The detailed explanation is good, but I think these lines are confusing.
* "blank must be a boolean, if true a postive number has a ' ' in place of the '-'." <- It's weird to say that a positive number has something in place of '-'. A positive number has no '-'. Perhaps it would be better to say that it adds an additional ' ' in front of the number, so that it is aligned with negative numbers with the same number of digits.
* plus must be a boolean, if true a '+' is used in place of the '-'. <- Same, but even more confusing. Naively, it sounds like we would format `-5` as `"+5"` when `plus` is true.

test_suite/format.jsonnet
```diff
@@ -53,6 +53,27 @@ std.assertEqual(std.format('thing-%-5.3d', [10.3]), 'thing-010  ') &&
 std.assertEqual(std.format('thing-%#-5.3d', [10.3]), 'thing-010  ') &&
 std.assertEqual(std.format('thing-%#-5.3i', [10.3]), 'thing-010  ') &&
 std.assertEqual(std.format('thing-%#-5.3u', [10.3]), 'thing-010  ') &&
+std.assertEqual(std.format('thing-%5.3d', [-10.3]), 'thing- -010') &&
```
sbarzowski, line 56: Nice to have all these cases covered!

PR #803, opened 2020-05-04
Fix clamp equivalent example in docs

JessThrysoee:


JessThrysoee:
@googlebot I signed it!

sbarzowski:
Oh, good catch. Thanks a lot.

Issue #196, opened 2016-05-26
Relax computed importstr ?


ant31:
Do you think it will be possible to use `importstr` with variable? 

e.g: 

``` yaml
 {
 files: [
        { 
         file: filepath,
         content: importstr filepath
         } for filepath in ['a.txt', 'b.txt', 'c.txt']
  ],
}
```

``` yaml
some_function(file):: (
    local content = importstr file;
    #process the content
  )
```

ant31:
Any technical reason for `importstr` to not allow import variable  ? 
I understand for `import`, but I think imporstr is a little different, it's equivalent to do an `file.open(filepath)` in other languages, and they mostly all accept a variable.

sparkprime:
Widening it to a variable is just as "bad" as widening it to an arbitrary expression, so if we were going to do anything it would be that.

The property that we gain by having the restriction is that you can do a simple static analysis and determine which files are dependencies, and then determine whether or not the code needs re-executing.  This is potentially useful although I'm not sure anyone has made use of it yet.  However if we lift the restriction and lose the property, we won't be getting it back again.  So it needs careful thought.

There is one difference between Jsonnet and other programming languages, and that's that everybody writes programs that terminate in a relatively short period of time, so one answer is that if you want to know what files affect the evaluation of a config, you should execute the config.  Or, in the case of deciding whether or not a cached JSON value needs recomputing, save not just the JSON output but the list of files used to compute it.

ant31:
Thank you for the detailed explanation .

> we won't be getting it back again 
What about an explicit keyword or function to express it? 
like:  importFrom(expression) , this way users explicitly chose to lose the static analysis property  ?

ant31:
I'm closing it, 
I've been able to do it easily via a native extension.

Thank you.

sparkprime:
I suppose native extensions broke static analysis already :)  Same as they do in Java etc.

nikolay:
@sparkprime The stdlib is so limited and so opinionated that, basically, we're forced to write native extensions, which kind of defeats the purpose of using Jsonnet.

sbarzowski:
@nikolay
It's not about the stdlib, it's about the language. Hermiticity is a core principle of Jsonnet. The output shouldn't be affected by the environment. Jsonnet code shouldn't jump around the filesystem and read files. Import mechanism is designed for importing Jsonnet libraries etc. and not for arbitrary IO.

[...]

PR #805, opened 2020-05-08
POC for #196 - Relax computed importstr

mbarbero:
Adds a new builtin function (importstrDyn) that accept a computed string.

I did not modified the importstr logic itself as I don't understand well
enough how the parser is working. Of course, it would better if importstr could be modified to accept any computed string.

With `test.jsonnet`

```
{
  a: {
    b: "dyn",
    c1: importstr "file-dyn.txt",
    c2: std.importstrDyn("file-dyn.txt"),
    c3: std.importstrDyn("file-%s.txt" % self.b),
  },
}
```

and with `file-dyn.txt`

```
>> 42
```

Executing `$ jsonnet test.jsonnet` outputs 

```
{
   "a": {
      "b": "dyn",
      "c1": ">> 42",
      "c2": ">> 42",
      "c3": ">> 42"
   }
}
```

Hopefully, this PR will trigger new discussions and maybe the addition of such a feature. I really don't see how this would diminish the hermiticity (not more than extVar or tla IMO) nor how it would prevents static analysis.

### DIFF
--- a/core/desugarer.cpp
+++ b/core/desugarer.cpp
@@ -34,7 +34,7 @@ struct BuiltinDecl {
     std::vector<UString> params;
 };
 
-static unsigned long max_builtin = 37;
+static unsigned long max_builtin = 38;
 BuiltinDecl jsonnet_builtin_decl(unsigned long builtin)
 {
     switch (builtin) {
@@ -76,6 +76,7 @@ BuiltinDecl jsonnet_builtin_decl(unsigned long builtin)
         case 35: return {U"parseJson", {U"str"}};
         case 36: return {U"encodeUTF8", {U"str"}};
         case 37: return {U"decodeUTF8", {U"arr"}};
+        case 38: return {U"importstrDyn", {U"str"}};
         default:
             std::cerr << "INTERNAL ERROR: Unrecognized builtin function: " << builtin << std::endl;
             std::abort();
--- a/core/vm.cpp
+++ b/core/vm.cpp
@@ -734,7 +734,7 @@ class Interpreter {
      */
     HeapThunk *import(const LocationRange &loc, const LiteralString *file)
     {
-        ImportCacheValue *input = importString(loc, file);
+        ImportCacheValue *input = importLiteralString(loc, file);
         if (input->thunk == nullptr) {
             Tokens tokens = jsonnet_lex(input->foundHere, input->content.c_str());
             AST *expr = jsonnet_parse(alloc, tokens);
@@ -756,11 +756,14 @@ class Interpreter {
      * \param file Path to the filename.
      * \param found_here If non-null, used to store the actual path of the file
      */
-    ImportCacheValue *importString(const LocationRange &loc, const LiteralString *file)
+    ImportCacheValue *importLiteralString(const LocationRange &loc, const LiteralString *file)
+    {
+        return importString(loc, file->value);
+    }
+
+    ImportCacheValue *importString(const LocationRange &loc, const UString &path)
     {
         std::string dir = dir_name(loc.file);
-
-        const UString &path = file->value;
 
         std::pair<std::string, UString> key(dir, path);
         ImportCacheValue *cached_value = cachedImports[key];
@@ -881,6 +884,7 @@ class Interpreter {
         builtins["parseJson"] = &Interpreter::builtinParseJson;
         builtins["encodeUTF8"] = &Interpreter::builtinEncodeUTF8;
         builtins["decodeUTF8"] = &Interpreter::builtinDecodeUTF8;
+        builtins["importstrDyn"] = &Interpreter::builtinImportstrDyn;
     }
 
     /** Clean up the heap, stack, stash, and builtin function ASTs. */
@@ -1188,6 +1192,15 @@ class Interpreter {
     {
         validateBuiltinArgs(loc, "log", args, {Value::NUMBER});
         scratch = makeNumberCheck(loc, std::log(args[0].v.d));
+        return nullptr;
+    }
+
+    const AST *builtinImportstrDyn(const LocationRange &loc, const std::vector<Value> &args)
+    {
+        validateBuiltinArgs(loc, "importstrDyn", args, {Value::STRING});
+        const auto *str = static_cast<const HeapString *>(args[0].v.h);
+        const ImportCacheValue *importedValue = importString(loc, str->value);
+        scratch = makeString(decode_utf8(importedValue->content));
         return nullptr;
     }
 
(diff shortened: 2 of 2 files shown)


mbarbero:
Any feedbacks?

ant31:
This is one of my main blocker to not create external commandline tool in order to use std.native function and do the same.

sbarzowski:
Thanks for the ping and sorry I didn't reply sooner.  This is a long standing discussion. Basically the import mechanism was intended as a way to statically import libraries and split code into multiple packages, not for loading stuff dynamically. The matter is not settled, but I expect the impact on the language to be (slightly) net negative.

Here's my summary of arguments for and against relaxing the restriction.

Pros:
* Allows creating functions like `importJson`, `importYaml`, `importIni`.
* Allows reading existing data represented as multiple files more easily.
* Allows using `import` and `importstr` on the same file without duplication (useful for documentation generation).
* Allows using plain data manifests to configure Jsonnet processing (see https://github.com/google/jsonnet/issues/196#issuecomment-356911064). 
* ?

Cons:
* No longer possible to statically determine all the paths that could be imported from a given file. This gets in the way of some current efforts (linter, documentation generator) and some potential ones too.
* Related to the previous point, it's impossible to audit all accessed files by simply grepping for import / importstr.
* Encourages programming patterns which make it harder to see where the code is coming from. The imports are intended mostly to be used in the same way as imports in Python, includes in C++ etc. That is, a bunch of statically known imports at the top which refer to other files in the same repository or to dependencies which are provided in JPATH. If we added computed import people would start introducing prefixes, functions which read a bunch of files based on conventions etc.
* Encourages calculating the absolute path instead of using JPATH for handling dependencies.
* Enables doing some wildly inappropriate things such as reading a manifest of installed Jsonnet configs from absolute path and importing Jsonnet files based on paths found there. 
* Makes it harder for people to migrate between different importers, because they depend more on the path structure.
* Makes it harder to add New Better Imports™ one day (e.g. dotted imports with better handling of dependencies-of-dependencies), because people do their own calculation.

The cons are not really deal breakers, but the pros seem like relatively minor convenience things and the alternatives are not so bad.

The most important things if you want to move this forward would be to describe the use cases and why the alternatives are not sufficient (why not specify the paths directly and why not pass the data as TLA).

FYI the implementation on the interpreter side is really not a problem. This is entirely an issue of language design.

> This is one of my main blocker to not create external commandline tool in order to use std.native function and do the same.

I'm not sure what you mean. Can you elaborate? Do you mean using Jsonnet as a library and providing a native function there which emulates computed imports? Do you still have the same use case as you mentioned in #196?

ant31:
> I'm not sure what you mean. Can you elaborate? Do you mean using Jsonnet as a library and providing a native function there which emulates computed imports? 

Yes, using Jsonnet as a library to implement additional methods such as `importJson`, `importYaml`.
Several other tools based on Jsonnet are doing the same like (bitnami/kubecfg)[https://github.com/bitnami/kubecfg/blob/master/lib/kubecfg.libsonnet#L19-L25]  or (grafana/tanka)[https://github.com/grafana/tanka/blob/e59cc470cc721f35110fa2537d19c273a944d3e6/docs/docs/jsonnet/native-functions.md#parsejson].

sbarzowski:
Yeah, importYaml stuff is is the most convincing use case here. We're also looking into alternative solutions, e.g. first-class support for loading different types of files with `import`.

And you can always do things like `std.parseJson(importst 'yourfile.json')` which is maybe slightly verbose, but it works.

ant31:
@sbarzowski also I like to 'hide' jsonnet from user when it's not needed and make it like seems like basic json/yaml. 

So instead to requires the end-user to know about jsonnet to write the following:
```
build(
{
config: std.parseJson(importstr './config/myconfig.json')
}
)
```
I prefer: 
```
build(
{
config_file:  './config/myconfig.json'
}
)
```
And process the user input in the `build` function.
This way, everything inside `build()` is very close to raw json. it is a good way to onboard new users to jsonnet and slowly ramp their knowledge about it.

mbarbero:
@sbarzowski Thanks for your effort in identifying our precise needs.

> > Phase 2 generates kubernetes secret resources by importstr the content of the secret files.
> 
> How about loading the content of secret files using extVar/tlaVar? Is there some reason why that wouldn't be practical?

We're talking dozens of files here. It would create humongous command line or function parameters list. Also, secrets file don't just contain a single password or passphrase. Sometimes, it's complete files with dozens of passwords in XML... (and this is out of our control). That makes both extVar and tlaVar impracticable.

ant31:
> Trying to hide Jsonnet from the users sounds is a bit of a boundary case from language design perspective. Are you sure importstr is not acceptable? You don't need to wrap it in parseJson in that part of code, so it's just a matter of importstr.

Not that much, it actually joins @mbarbero  case, because you could externalise this configuration/user input in a yaml :

values.yaml:
```yaml
config_file: ./config/myconfig.json
username: me
password: bla
```

manifest.jsonnet
```jsonnet
{
values: parseYaml(importstr 'config.yaml')
config: importDyn(values.config_file)
}
```

sbarzowski:
You don't need to pass every single secret as a separate tla/extVar – that of course could get quickly out of hand. I meant passing all secrets at once in one structure file. E.g. you could have a secrets.json which contains all secrets (or it could contain XMLs which contains the actual if that's your thing). 

If you already have a directory structure with all these secrets, you can serialize them all to JSON with a tool such as `dir2json` (https://github.com/sbarzowski/dir2json).

```
# I imagine it could be the last step in the "imperative stage"
dir2json generated_secrets/ > secrets.json
jsonnet --ext-var secrets=secrets.json foo.jsonnet
```

The main disadvantage of this approach is having to copy all of these secrets to one place ahead of time, whether or not they're going to actually be used. It also doesn't work well if your directory structure has symlinks etc. This is something you can use here and now, independently from what we decide here.

sbarzowski:
What do you think about providing `--tla-var-dir`/`--ext-var-dir` which would provide an object similar to what you would get with `dir2json` but built more efficiently and without the need of external tool and temporary files. Perhaps it could even load the files lazily. It would load all files inside as strings, so it wouldn't allow running Jsonnet files from inside. 

How useful do you think that would be for use cases like yours?

sbarzowski:
@sparkprime Since we got back to this discussion, perhaps you have some thoughts?

mbarbero:
> What do you think about providing `--tla-var-dir`/`--ext-var-dir` which would provide an object similar to what you would get with `dir2json` but built more efficiently and without the need of external tool and temporary files. Perhaps it could even load the files lazily. It would load all files inside as strings, so it wouldn't allow running Jsonnet files from inside.
> 
> How useful do you think that would be for use cases like yours?

@sbarzowski Sorry for late reply. Yes, this solution would be workable. As long as it's supported natively by jsonnet (without having to call the intermediate dir2json binary), I think it's a decent alternative to computed `importstr` in my case.

MrG9090 (approved):

MrG9090 (approved):

Issue #804, opened 2020-05-07
Questions and errors in the spec


Alxandr:
According to the language spec, the following is the ways you can slice an array:

![image](https://user-images.githubusercontent.com/112334/81320112-f5197c80-9090-11ea-8a18-f355f44fa9b9.png)

As far as I can tell, this does not accept the following strings:

* `foo[::]`
* `foo[1::]`
* `foo[::1]`
* `foo[1::1]`

However, as from what I can see, all of those works. Is this just an error in the BNF? Or am I reading it wrong?

sbarzowski:
Thanks for catching this. I think you're right. The second and third expr should be marked as optional. And I think the second rule is going to be redundant then.

We support all the variants you mentioned. We even have rewriting rules for desugaring them in the next section of the spec:
![image](https://user-images.githubusercontent.com/683828/81324591-7b38c180-9097-11ea-8de0-5c32512a9ab5.png)

@sparkprime Is that correct?

@Alxandr If you feel like contributing to Jsonnet documentation, this snippet comes from here: https://github.com/google/jsonnet/blob/v0.15.0/doc/ref/spec.html#L328. That's completely optional ofc, if not, I'll do it myself later.

Alxandr:
I'm currently busy porting the whole thing to rust (for fun), so maybe after xD. Gonna leave this issue in my github inbox so I remember it til then, but it will probably be a week at least.

Also; while we're on the topic: `{ [x]+:true for x in ['a'] }` is not legal according to the BNF, but is accepted by the parser. I've not gotten far enough yet to figure out whether this should be legal or not.

sbarzowski:
That's cool! If you have any questions feel free to ask them here. Please report any discrepancies/bugs that you find – reimplementation allows to catch very tricky issues and it's a great way for us to make sure that docs, tests and code stay in sync.

I think you're also right about the `+:` in comprehensions and I think it should be legal. It's a bit weird thing to do, but it's clear what it should mean. I'm not sure if it's covered by the test suite or if anyone does that in the real code.

Alxandr:
It would be great if someone could say definitely whether or not it should be legal or not. Currently, I've set up my parser not to accept it, but it's easily enough to reverse. On the same note, the fact that you can't use `::` and `:::` in comprehension I cannot find anywhere in the spec. Nor have I yet figured out what `:::` even means (but I've only read the parsing section yet, with some minor exceptions).

[...]

Issue #343, opened 2017-08-14
document tailstrict


guoshimin:
Include tailstrict in the tutorial and language specification.

sparkprime:
example to use in docs:

```
local sum_from_one_to_n(counter, n) =
  if n == 0 then
    counter
  else
    sum_from_one_to_n(counter + n, n - 1) tailstrict;

sum_from_one_to_n(0, 10)
```

sbarzowski:
I needed to dive a little bit deeper into this (b/c of Go port).

So `tailstrict` actually does two things:
1) Strictness. It forces evaluation of arguments, before the body of function is executed. This part actually affects semantics of the language:
```
local foo(x) = 42; foo(error "xxx")
> 42
```

```
local foo(x) = 42; foo(error "xxx") tailstrict
> RUNTIME ERROR: xxx
```
or perhaps more importantly:
```
local foo(x) = 42; foo(endlessLoop()) tailstrict
> <hangs>
```

2) Tail call optimization of the callee. This part only affects efficiency and stack traces. When a called function makes a tail call, we remove the callee from the stack.

Example:
```
local foo(x) = if x == 0 then error "err" else foo(x - 1);
local bar(x) = foo(x) tailstrict;
bar(2)
```
Result:
```
RUNTIME ERROR: err
        tail2.jsonnet:1:31-41   function <foo>
        tail2.jsonnet:1:48-57   function <foo>
        tail2.jsonnet:1:48-57   function <bar>
        tail2.jsonnet:3:1-6
```

Example without tailstrict:
```
local foo(x) = if x == 0 then error "err" else foo(x - 1);
local bar(x) = foo(x);
bar(2)
```
Result:
```
RUNTIME ERROR: err
        tail.jsonnet:1:31-41    function <foo>
        tail.jsonnet:1:48-57    function <foo>
        tail.jsonnet:1:48-57    function <foo>
        tail.jsonnet:2:16-21    function <bar>
        tail.jsonnet:3:1-6
```

This is quite a bit different from what I expected at first. In particular:
- It's the callee and not the caller that gets removed from stack.
- It doesn't matter at all if a call marked tailstrict is a tail call. It's tail call optimization in the callee, not here. And in the callee it automatically recognizes tail calls.
- (2) by itself would be safe to use on each call. (1) + (2) is safe to use on each call that will evaluate all arguments anyway. 

[...]

PR #806, opened 2020-05-21
Fix spec errors pointed out in #804.

sbarzowski:
This change only covers clear, simple errors
and not stylistic issues or spec-implementation
discrepancies.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -35,6 +35,7 @@ bazel-*
 /external/
 /dist/
 /jsonnet.egg-info/
+/doc/.jekyll-cache/
 
 # Cmake
 **/CMakeCache.txt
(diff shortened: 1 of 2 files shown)


sbarzowski:
@sparkprime @Alxandr

Alxandr:
Reading latex isn't my strong suit, but I can point out one more thing while I'm at it. As far as I can see, the spec currently does not allow for `tailstrict` anywhere (nor does it describe what it does).

sbarzowski:
Yeah, it's high time to document it. It's tracked in #343 (which I've just "upgraded" to a bug) and you can find some explanation there.

Alxandr:
Welp. That did not do what I expected it to do. Glad I found this info before I got to implementing it 😛.

PR #816, opened 2020-06-01
[Docs] Generate stdlib docs with Jsonnet.

sbarzowski:
Benefits:
* Keeping things consistent should be easier.
* We can now change the way it is represented in HTML
  without going over the whole thing.
* We can now use the stdlib data programatically, so
  it should be easier to check it's in line with reality.
* Make it easier to integrate with any potential
  documentation generation tools.
* First step towards completely Jsonnet-based
  site generation which I think is worth exploring.


doc/_stdlib_gen/stdlib-content.jsonnet
```diff
@@ -0,0 +1,1209 @@
+local html = import 'html.libsonnet';
```
sparkprime, line 1: How do you feel about this file being pure data and then having it exported from comments in the code?

sparkprime:
Nice.  I suppose you could have generated XML too but this is going to be easier.  Being able to check that the set of documented functions is the same as the set of actual functions in stdlib is very nice.  I guess we could do that before with some grep hackery but now it's easy.

What's the status of doc generation?  I'd love to do the analysis part.  Schools have just started up again so I might actually be able to do it now

doc/ref/stdlib.html
```diff
@@ -1,12 +1,15 @@
 ---
+# AUTOGENERATED FILE. DO NOT EDIT BY HAND!
 layout: default
 title: Standard Library
 ---
 
 <div class="hgroup">
   <div class="hgroup-inline">
     <div class="panel">
-      <h1 id="standard_library">Standard Library</h1>
```
sparkprime, line 9: It's missing the id.  Guess we don't need it though.

doc/ref/stdlib.html
```diff
@@ -78,61 +84,64 @@ <h3 id="types_reflection">Types and Reflection</h3>
...
   </div>
 </div>
-
 <div class="hgroup">
   <div class="hgroup-inline">
     <div class="panel">
       <p>
         Return a string that indicates the type of the value. The possible return values are:
-        "array", "boolean", "function", "null", "number", "object", and "string".
+        "array", "boolean", "function", "null", "number", "object", and "string"."
```
sparkprime, line 121: Extra quote looks dubious

doc/ref/stdlib.html
```diff
@@ -1037,25 +1223,24 @@ <h3 id="manifestation">Manifestation</h3>
...
-      </p>
-
-
-      <pre>{
+    Convert the given structure to a string in <a href="http://en.wikipedia.org/wiki/INI_file">INI format</a>. This
+    allows using Jsonnet's
+    object model to build a configuration to be consumed by an application expecting an INI
+    file. The data is in the form of a set of sections, each containing a key/value mapping.
+    These examples should make it clear:
+</p>
```
sparkprime, line 1242: It's suddenly got broken indentation here, but was working fine further up the file.  Intended?

doc/ref/stdlib.html
```diff
@@ -1902,52 +2179,59 @@ <h4 id="setUnion">std.setUnion(a, b, keyF=id)</h4>
...
-        <code>[1, 2, 3]</code>.
+        <p>
+          Example: <code>std.setUnion([1, 2], [2, 3])</code> yields <code>[ 1, 2, 3 ]</code>.
+        </p>
       </p>
       <p>
-        Example2: <code>std.setUnion([{n:"A", v:1}, {n:"B"}], [{n:"A", v: 9999}, {n:"C"}],keyF=function(x) x.n)</code>
-        yields
-        <code>[{n: "A", v:1}, {n: "B"}, {n: "C"}]</code>.
+        <p>
```
sparkprime, line 2187: Extra <p> here

sparkprime: Extra `<p>` here

doc/_stdlib_gen/stdlib-content.jsonnet
sbarzowski: Yeah, I think this is the right way. 

In a sense this is pure data already – there are no functions or hidden fields in the output object. Any function calls here are just for readability convenience. In other words you can evaluate it and serialize it to JSON and use that instead of using `stdlib-content.jsonnet` directly.

sbarzowski:
> What's the status of doc generation? I'd love to do the analysis part. Schools have just started up again so I might actually be able to do it now

I haven't looked into the analysis part. Please go ahead. And if you want to discuss something ping me anytime.

There's also more work needed on representation (e.g. how to describe objects in JSON) and rendering (e.g. nicely formatting examples and generally cleaner html) before it can before we can present it as a general purpose tool. I plan to do these things in future PRs.

doc/ref/stdlib.html
sbarzowski: I've added it. The intention in this change is to keep it the same modulo formatting, obvious typos and such.

doc/ref/stdlib.html
sbarzowski: Fixed.

doc/ref/stdlib.html
sbarzowski: The whole description was marked as verbatim html, so that <pre> wouldn't get messed up. This was intentional, but I've added a proper support for pre now.

doc/ref/stdlib.html
sbarzowski: Fixed.

PR #817, opened 2020-06-03
[Website] Fix docs for std.rstripChars

camh-:
From what looks like to be a cut-and-paste error, `std.rstripChars` is
documented as removing chars from the beginning and end of a string,
which should just be the end.


sbarzowski:
Yes, you're right. Thanks for fixing this.

PR #844, opened 2020-09-17
introduce idLocation field to the ObjectField

Dobryk15:
Add idLocation to store the location of the object field name when this name is an identifier (see example below).
Then, the location info will be used in type error message to refer to the exact place in the program where the error occurred.
Type errors occur when the type inference algorithm fails due to type mismatch in the program. 
Type inference for Jsonnet is my intern project at Google (see https://github.com/Dobryk15/jsonnet). 

Example:
```
{
    x: 1
}
```
Before suggested changes, the location of _x_ was undefined and represented in desugared AST as Location("0:0,0:0").
After applied changes, the location will be Location("2:5,2:6").
As a result, there is location info for a field name written not only as a string ('x') but also as an identifier (x).

### DIFF
--- a/core/ast.h
+++ b/core/ast.h
@@ -638,6 +638,7 @@ struct ObjectField {
     bool methodSugar;  // f(x, y, z): ...  (ignore if kind  == ASSERT)
     AST *expr1;        // Not in scope of the object
     const Identifier *id;
+    LocationRange idLocation;
     ArgParams params;    // If methodSugar == true then holds the params.
     bool trailingComma;  // If methodSugar == true then remembers the trailing comma.
     Fodder opFodder;     // Before the : or =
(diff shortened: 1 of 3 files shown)


Dobryk15:
@googlebot I signed it!

Dobryk15:
@googlebot I signed it!

sbarzowski:
Looks good, thanks! 

BTW feel free to message me internally. I'm curious about your project and I'll be happy to help.

Dobryk15:
@sbarzowski Thank you for offering help! If there will be any questions - I will ask.
And I am happy to hear that this project sounds interesting to you. I or my host can also share design document with you and at the end of the internship, our final results.

sbarzowski:
> I or my host can also share design document with you

Sounds good, please share it with me.

Dobryk15:
Ok. I still need to complete some parts of the document and will share with you within next week.

PR #856, opened 2020-10-19
fix cleaning important data.

ihsinme:
in your project in the md5.cpp file on line 245, data cleansing is present. The specified procedure can be optimized by the compiler according to
```
Notes
memset may be optimized away (under the as-if rules) if the object modified by this function is not accessed again for the rest of its lifetime (e.g. gcc bug 8537). For that reason, this function cannot be used to scrub memory (e.g. to fill an array that stored a password with zeroes). This optimization is prohibited for memset_s: it is guaranteed to perform the memory write. Third-party solutions for that include FreeBSD explicit_bzero or Microsoft SecureZeroMemory.
```
https://en.cppreference.com/w/c/string/byte/memset

with that in mind, I suggest adding the -fno-builtin-memset flag to eliminate this.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -31,7 +31,7 @@ OPT ?= -O3
 
 PREFIX ?= /usr/local
 
-CXXFLAGS ?= -g $(OPT) -Wall -Wextra -Woverloaded-virtual -pedantic -std=c++0x -fPIC
+CXXFLAGS ?= -g $(OPT) -Wall -Wextra -Woverloaded-virtual -pedantic -std=c++0x -fPIC -fno-builtin-memset
 CXXFLAGS += -Iinclude -Ithird_party/md5 -Ithird_party/json
 CFLAGS ?= -g $(OPT) -Wall -Wextra -pedantic -std=c99 -fPIC
 CFLAGS += -Iinclude


ihsinme:
@googlebot I signed it!

sbarzowski:
Thanks for sending the PR. That's correct, but I'm not sure if it's a problem in our case. We already have GC and make no guarantees about when the input strings will actually be freed (and they won't be zeroed before) before evaluation finishes. So, I'm afraid taking care of intermediate values doesn't make it any more secure. 

If your attack model includes the baddies looking at deallocated memory before it's reclaimed by the OS or overwritten, I would stick to languages with more direct control over memory (such as C).

ihsinme:
Good afternoon.
thanks for the answer, maybe I misunderstood it.
What I'm talking about is that the compiler will most likely remove your memset. if you think that this deletion does not affect anything, then delete it yourself.
otherwise you will confuse the user.

The threat vector that can be considered is quite general and falls under https://cwe.mitre.org/data/definitions/14.html

sbarzowski:
> What I'm talking about is that the compiler will most likely remove your memset. if you think that this deletion does not affect anything, then delete it yourself.

Yes, I think this battle is already lost and memset could as well be removed. If you process sensitive data with Jsonnet, is may linger in memory for a while and we can't do much about it in general. 

> if you think that this deletion does not affect anything, then delete it yourself.

I agree with your reasoning. If it was "our" code I would remove the memset. However, the md5 code is a vendored 3rd party library, so I prefer not to touch it unless something is actually broken.

ihsinme:
Thanks.
Now everything is clear.

Should I close this PR?

sbarzowski:
Thank you for bringing this to us. Please let us know if you notice any other potential security issue.

PR #857, opened 2020-10-22
docs: add std.slice documentation

### LINKED ISSUE #798: std.slice missed in documentation
`std.slice()` function is not described in the documentation for the standard library.
https://github.com/google/jsonnet/blob/master/stdlib/std.jsonnet#L191
https://jsonnet.org/ref/stdlib.html

zam-zam:
This PR fixes #798

### DIFF
--- a/doc/_stdlib_gen/stdlib-content.jsonnet
+++ b/doc/_stdlib_gen/stdlib-content.jsonnet
@@ -1003,13 +1003,22 @@ local html = import 'html.libsonnet';
         {
           name: 'slice',
           params: ['indexable', 'index', 'end', 'step'],
-          description: |||
-            Selects the elements of an array or a string from <code>index</code> to <code>end</code> with <code>step</code> and returns an array or a string respectively.
-          |||,
+          description: html.paragraphs([
+            |||
+              Selects the elements of an array or a string from <code>index</code> to <code>end</code> with <code>step</code> and returns an array or a string respectively.
+            |||,
+            |||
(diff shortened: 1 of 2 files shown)


zam-zam:
@googlebot I fixed it.

doc/_stdlib_gen/stdlib-content.jsonnet
```diff
@@ -1000,6 +1000,23 @@ local html = import 'html.libsonnet';
             },
           ],
         },
+        {
+          name: 'slice',
+          params: ['indexable', 'index', 'end', 'step'],
+          description: |||
+            Selects the elements of an array or a string from <code>index</code> to <code>end</code> with <code>step</code> and returns an array or a string respectively.
```
sbarzowski, line 1007: Please add a note that it's recommended to use dedicated slicing syntax (e.g. arr[2:4] instead of std.slice(arr, 2, 4, 1)).

doc/_stdlib_gen/stdlib-content.jsonnet
```diff
@@ -1000,6 +1000,23 @@ local html = import 'html.libsonnet';
...
+          |||,
+          examples: [
+            {
+              input: 'std.slice([1, 2, 3, 4, 5], 1, 3, 1)',
+              output: std.slice([1, 2, 3, 4, 5], 1, 3, 1),
+            },
+            {
+              input: 'std.slice("jsonnet", 0, 4, 1)',
```
sbarzowski, line 1015: Please add an example with `step` != 1. This is probably the least obvious parameter, so we should show what it does.

sbarzowski:
Thanks! We should mention the slicing syntax and add an example with step. Otherwise looks good.

Issue #543, opened 2018-07-17
iterating through objects as key value pairs (cf. pythons .items())


lukasheinrich:
sometimes it's useful to process objects by iterating through the key value pairs

python has 

```
obj = {"a": 1, "b": 2}
{k:v*2 for k,v in obj.items()
```

jq has

```
echo '{"a": 1, "b": 2}'|jq '[to_entries[]|{key: .key, value: (.value * 2)}]|from_entries'
```

is there something similar in jsonnet? it seems like the list and object comprehensions could return (k,v) tuples if the right hand side is an object

sparkprime:
You can do `{k: obj[k]*2 for k in std.objectFields(obj)}`

Iteration over objects would be nice too `{k: e for k, v in ...}`
And it's just syntax sugar to implement it.

buremba:
@sparkprime The syntax sugar that you gave as an example is not implemented yet, right? It throws the following exception: `object comprehensions can only have [e] fields.`

buremba:
It looks like the syntax is as follows:`{ [key]: last_values[key] for key in std.objectFields(last_values) ` by looking at `std.jsonnet`. It would be great if we can add it to the documentation.

sparkprime:
The issue is about matching the python feature of for k, v in obj, which I think would be useful even though a simple work around is to loop over the fields doing obj[k] each time.  It is indeed not implemented yet.

doronbehar:
I also noticed the lack of such a function, and I was surprised that `std.mapWithKey` didn't generate an array. I opened https://stackoverflow.com/questions/63915576/jsonnet-std-mapwithkey-not-generating-an-array/63915745#63915745 and answered my self based upon @sbarzowski's response.

Also, I noticed there's a lack of `std.objectValues` which is a bit annoying as well, but once again, it can be worked around if using a `std.map` and `std.objetKeys`.

alexei-matveev:
The order of [k , v] pairs could be a problem for stable generated configs. FWIW, Clojure implements "Sequential Interface" for Maps/Dictionaries/Objects too, the order is also not guaranteed, of course...

PR #858, opened 2020-10-23
Implement std.objectValues and std.objectValuesAll

tragiclifestories:
Hello friends!

This PR adds two helper functions for extracting an array of values from an object. This was not impossible before (hence the implementations are in jsonnet), but I think it is worth having a 'first class' implementation - for further evidence, it is a [very](https://lodash.com/docs/4.17.15#values) [common](https://ramdajs.com/docs/#values) [feature](https://clojuredocs.org/clojure.core/vals) of data transformation libraries.

It was mentioned in [this comment](https://github.com/google/jsonnet/issues/543#issuecomment-693248735).

### DIFF
--- a/doc/ref/stdlib.html
+++ b/doc/ref/stdlib.html
@@ -207,6 +207,28 @@ title: Standard Library
 <div class="hgroup">
   <div class="hgroup-inline">
     <div class="panel">
+      <h4 id="objectValues">
+        std.objectValues(o)
+      </h4>
+    </div>
+    <div style="clear: both"></div>
+  </div>
+</div>
+<div class="hgroup">
+  <div class="hgroup-inline">
+    <div class="panel">
+      <p>
+        Returns an array of the values in the given object. Does not include hidden fields.
+      </p>
+      
+    </div>
+    <div style="clear: both"></div>
+  </div>
+</div>
+
+<div class="hgroup">
+  <div class="hgroup-inline">
+    <div class="panel">
       <h4 id="objectHasAll">
         std.objectHasAll(o, f)
       </h4>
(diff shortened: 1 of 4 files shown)


doc/ref/stdlib.html
```diff
@@ -63,7 +63,7 @@ <h4 id="extVar">
       <p>
         If an external variable with the given name was defined, return its string value. Otherwise, raise an error.
       </p>
-      
+
```
tragiclifestories, line 66: Sorry about this noise - I have an vim trigger thing that deletes trailing whitespace ...

sbarzowski: Please don't edit this file manually. Here are the instructions for updating stdlib docs: https://github.com/google/jsonnet#locally-serving-the-website.

sbarzowski (commented):
Thanks! This is a good addition to the standard library.

doc/ref/stdlib.html
tragiclifestories: Ah, that makes a lot more sense - updated.

stdlib/std.jsonnet
```diff
@@ -39,7 +39,7 @@ limitations under the License.
     assert std.isString(str) : 'substr first parameter should be a string, got ' + std.type(str);
     assert std.isNumber(from) : 'substr second parameter should be a string, got ' + std.type(from);
     assert std.isNumber(len) : 'substr third parameter should be a string, got ' + std.type(len);
-    assert len >=0 : 'substr third parameter should be greater than zero, got ' + len;
+    assert len >= 0 : 'substr third parameter should be greater than zero, got ' + len;
```
tragiclifestories, line 42: These changes are also all autoformatting but they look like good changes? I can rebase and just take the actual stuff I added if you want.

sbarzowski: Yeah, it looks like our own auto-formatting. Ideally that would be a separate commit, but that's not a big deal.

PR #859, opened 2020-10-28
libjsonnet: Make jsonnet_evaluate_*_stream respect jsonnet_string_output 

hanazuki:
The current implementation of jsonnet_evaluate_(snippet|file)_stream returns a list of JSON-encoded strings, whether the string_output option is set or not.

This patch corrects this behavior so that the functions return plain strings when the string_output option is set.


sbarzowski:
Thanks for fixing this!

PR #871, opened 2020-12-27
Add std. manifestJsonMinified

hanyucui:
This also partially implements the feature requested in https://github.com/google/go-jsonnet/issues/471

### DIFF
--- a/doc/_stdlib_gen/stdlib-content.jsonnet
+++ b/doc/_stdlib_gen/stdlib-content.jsonnet
@@ -693,78 +693,68 @@ local html = import 'html.libsonnet';
                 to break long lines. <code>key_val_sep</code> is used to separate the key and value
                 of an object field:
             |||),
-            html.pre({}, |||
-              std.manifestJsonEx(
+          ],
+          examples: [
+            {
+              input: |||
+                std.manifestJsonEx(
                 {
                     x: [1, 2, 3, true, false, null,
                         "string\nstring"],
                     y: { a: 1, b: 2, c: [1, 2] },
                 }, "    ")
-            |||),
-            html.p({}, |||
-              Yields a string containing this JSON object:
-            |||),
-            html.pre({}, |||
-              {
-                  "x": [
-                      1,
-                      2,
-                      3,
-                      true,
-                      false,
-                      null,
-                      "string\nstring"
-                  ],
-                  "y": {
-                      "a": 1,
-                      "b": 2,
-                      "c": [
-                          1,
-                          2
-                      ]
-                  }
-              }
-            |||),
-            html.p({}, |||
-              And here is an example that sets all arguments:
-            |||),
-            html.pre({}, |||
-              std.manifestJsonEx(
+              |||,
+              output:
+                std.manifestJsonEx(
+                {
+                    x: [1, 2, 3, true, false, null,
+                        "string\nstring"],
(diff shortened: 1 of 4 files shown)


hanyucui:
@sbarzowski Hi, this PR is still WIP, but could you kindly take a look and let me know if it's in the right direction?

sbarzowski:
Thanks. It looks good.

Two nits:
* I would use more specific arguments in manifestJsonEx instead of a single `minimized` boolean. Maybe something like `newline` `and `colon_whitespace`. We can also consider something more abstract like `item_separator_whitespace` instead of `newline` (because for example a single space would also make sense).
* s/minimized/minified/ in the function name. I know minimized is also used in this sense, but minified seems more precise.

We'll need tests and documentation changes (https://github.com/google/jsonnet/blob/v0.17.0/doc/_stdlib_gen/stdlib-content.jsonnet. Of course, changes to the native implementation in Go would be nice to have, but we can finish it here first.

doc/_stdlib_gen/stdlib-content.jsonnet
```diff
@@ -684,11 +684,14 @@ local html = import 'html.libsonnet';
         },
         {
           name: 'manifestJsonEx',
-          params: ['value', 'indent'],
+          params: ['value', 'indent', 'newline', 'key_val_sep'],
```
sbarzowski, line 687: Can you add `availableSince: 'upcoming'`? 

(See `manifestTomlEx`).

doc/_stdlib_gen/stdlib-content.jsonnet
```diff
@@ -722,6 +725,46 @@ local html = import 'html.libsonnet';
                   }
               }
             |||),
+            html.p({}, |||
```
sbarzowski, line 728: Any reason to put examples in the description instead of using `examples` field? The latter is generally better, because it allows us to change the presentation, e.g. we want to make the examples interactive one day. 

See: `manifestTomlEx`.

FYI The only reason `manifestXmlJsonml` and some others keep everything in the description is that they have some elements which are not supported properly yet (e.g. an svg picture illustrating an example).

We should add some documentation about this file, sorry for the burden of figuring it out based on examples alone.

sbarzowski (approved):
Nice!

doc/_stdlib_gen/stdlib-content.jsonnet
hanyucui: @sbarzowski I took a second look, it seems the `examples` field does not preserve newlines. The example for `manifestTomlEx` is flattened a single line and seems hard to read. What are your thoughts?

doc/_stdlib_gen/stdlib-content.jsonnet
hanyucui: @sbarzowski Should this be added to `manifestJsonMinified` instead? `manifestJsonEx` is available in the current version and I didn't find a field that indicates parameter list changes only.

sbarzowski: Yes, yes, that's what I meant.

doc/_stdlib_gen/stdlib-content.jsonnet
sbarzowski: Please leave it in `examples` and I'll fix the formatting later.

PR #875, opened 2021-01-04
Fix error messages for std.min

mhutter:
Currently, the error message incorrectly states that an error occurred in "std.max".


sbarzowski:
Thanks!

PR #876, opened 2021-01-06
fix optimized build

mikedanese:
Build is failing because we are using assert which is optimized away in
NDEBUG builds. This causes the a maybe-uninitialized warning, which in
turn causes the build to fail when used with -Werror.

Warnings:

```
core/vm.cpp: In member function 'void {anonymous}::Interpreter::evaluate(const AST*, unsigned int)':
core/vm.cpp:2369:44: warning: 'func' may be used uninitialized in this function [-Wmaybe-uninitialized]
 2369 |                                 HeapThunk *func;
      |                                            ^~~~
core/vm.cpp: In member function 'void {anonymous}::Interpreter::evaluate(const AST*, unsigned int)':
core/vm.cpp:2369:44: warning: 'func' may be used uninitialized in this function [-Wmaybe-uninitialized]
 2369 |                                 HeapThunk *func;
      |                                            ^~~~
core/vm.cpp: In member function 'void {anonymous}::Interpreter::evaluate(const AST*, unsigned int)':
core/vm.cpp:2369:44: warning: 'func' may be used uninitialized in this function [-Wmaybe-uninitialized]
 2369 |                                 HeapThunk *func;
      |
```

### DIFF
--- a/core/vm.cpp
+++ b/core/vm.cpp
@@ -31,6 +31,18 @@ limitations under the License.
 #include "string_utils.h"
 #include "vm.h"
 
+/** Macro that dumps an error and aborts the program regardless of
+ *  whether NDEBUG is defined. This should be used to mark codepaths
+ *  as unreachable.
+ */
+#define JSONNET_UNREACHABLE()                                      \
+  do {                                                             \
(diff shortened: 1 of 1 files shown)


sbarzowski:
Thanks!

PR #881, opened 2021-01-12
Fix std.format carry-over to wholes.

### LINKED ISSUE #879: String formatting of floats gives incorrect value when rounding up fractions to wholes
When I format a float with `%.1f` I would expect the behaviour to be the same as in Python. For example, in Python:

```python
>>> '%.1f' % 0.95555
'1.0'
```

```python3
>>> '{:.1f}'.format(0.95555)
'1.0'
```

However, when the number would be '1.0', running this in jsonnet I get:

```
> jsonnet <(echo "'%.1f' % 0.95555") 
"0.10"
```

This happens when rounding up to other whole numbers too:

```
jsonnet <(echo "'%.1f' % 1.95555") 
"1.10"
```

Values where this would result in a number below one seem fine:

```
> jsonnet <(echo "'%.1f' % 0.85555") 
"0.9"
```

Values with more than one decimal place also seem fine:

```
> jsonnet <(echo "'%.2f' % 0.95555") 
"0.96"
```

sbarzowski:
E.g. '%.1f' % 0.99 used to be formated as '0.10'.

Fixes #879.

### DIFF
--- a/stdlib/std.jsonnet
+++ b/stdlib/std.jsonnet
@@ -548,13 +548,20 @@ limitations under the License.
     local render_float_dec(n__, zero_pad, blank, plus, ensure_pt, trailing, prec) =
       local n_ = std.abs(n__);
       local whole = std.floor(n_);
+      // Represent the rounded number as an integer * 1/10**prec.
+      // Note that it can also be equal to 10**prec and we'll need to carry
+      // over to the wholes.  We operate on the absolute numbers, so that we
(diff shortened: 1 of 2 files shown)


sbarzowski:
@benk-gc Let me know if you can think of any additional test cases worth checking. If you could take a look at the code and see if the new approach looks solid to you, that would also be great.

benk-gc:
@sbarzowski Looks good to me 👍

PR #883, opened 2021-01-18
Fix std.trace()'s documentation

hanyucui:


sbarzowski:
Thank you. Test failure ofc has nothing to do with this change. It's just that we were testing against pretty old GCC and our dependency reasonably decided not to support that anymore. I'll just bump the GCC version and then I will merge this.

leo-baltus:
It seems https://jsonnet.org/ref/stdlib.html is not updated yet.

PR #885, opened 2021-01-19
Add a missing header.

sbarzowski:
Some users reported problems with compiling the project. It seems like on some platforms we need to explicitly include a header which is not needed for the ones we test on. I have no way to reproduce, but hopefully that will fix it.

Thanks to @prazek for the tip about what might be wrong.

### DIFF
--- a/core/formatter.cpp
+++ b/core/formatter.cpp
@@ -17,6 +17,7 @@ limitations under the License.
 #include <algorithm>
 #include <set>
 #include <typeinfo>
+#include <uchar.h>
 
 #include "formatter.h"
 #include "lexer.h"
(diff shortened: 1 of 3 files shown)


sbarzowski:
Eh, apparently some setups don't support uchar.h header (but worked without it). Annoying. Not sure what to do here.

prazek:
Hmm, uchar.h seems to be header for C (https://en.cppreference.com/w/c/string/multibyte/char32_t) defining char32_t macro, while in C++11 it is a keyword: https://en.cppreference.com/w/cpp/keyword/char32_t
The compiler in which the problem was reported is a bit old (gcc 5.4.0 was realeased in mid 2016). Maybe there lies the problem?

PR #893, opened 2021-02-12
Correct std.splitLimit example

jsmcnair:


jsmcnair:
@googlebot I signed it!

sbarzowski:
Thanks for catching it!

(FYI to actually regenerate the website you need to run `./tools/scripts/update_web_content.sh`. This is not really important, because I need to run a script anyway to push the website).

Issue #897, opened 2021-03-11
Incorrect decoding of JSON strings containing non-BMP characters encoded with \u UTF16 escapes


rhowe:
It looks like jsonnet is failing to parse JSON strings containing embedded \u escape sequences.

[RFC4627 section 2.5](https://tools.ietf.org/html/rfc4627#section-2.5) says that characters outside the Basic Multilingual Plane (BMP) can be encoded either in the native encoding of the JSON file itself or as its UTF-16 surrogate pair with each pair of bytes escaped with `\u`. This was news to me.

If my interpretation of the RFC is correct, these two UTF-8 JSON files should be equivalent:

test1.json:
```
["\uD83D\uDCD8"]
```
and test2.json (as generated by `base64 -d > test2.json <<<"WyLwn5OYIl0K"`):
```
["📘"]
```

I therefore expected the following jsonnet program to decode them similarly, but it doesn't:

```
{
  "test1.json": import 'test1.json',
  "test2.json": import 'test2.json',
}
```
Using jsonnet 0.17.0, the blue book in test1.json is encoded as `0xed 0xa0 0xbd 0xed 0xb3 0x98` in the output whereas from test2.json it is encoded as UTF-8, i.e. `0xf0 0x9f 0x93 0x98`

`0xed 0xa0 0xbd` is the UTF-8 encoding for U+D83D and `0xed 0xb3 0x98` is the UTF-8 encoding for U+DCD8 so I think these non-BMP characters are being treated as though they are BMP characters.

I think the issue is here: https://github.com/google/jsonnet/blob/master/core/string_utils.cpp#L114. Is the solution to validate whether the codepoint is a BMP one or not, and if not, attempt to parse a subsequent `\u` escape, then decode the result of the two as UTF-16?

sbarzowski:
Thanks you very much for reporting this. It seems that you're right. In particular, I was able to get the behavior you expected when parsing JSON in Python and with `std.parseJson` in Jsonnet (which uses a faster 3rd party parser). So it's a bug in our parser and we should fix this.

rhowe:
Thanks for responding - I didn't realise `std.parseJson` used a different parser and indeed you're right - it seems to deal with these escapes correctly, as demonstrated by this test:

```
{
  "test1.json (import)": import 'test1.json',
  "test1.json (importstr+parseJson)": std.parseJson(importstr 'test1.json'),
  "test2.json (import)": import 'test2.json',
[...]

PR #898, opened 2021-03-11
Handle JSON strings containing non-BMP characters

rhowe:
This extends the JSON string parsing to handle Unicode characters outside the Basic Multilingual Plane when they are represented using two UTF-16 surrogate pairs encoded using `\u` escapes.

Fixes https://github.com/google/jsonnet/issues/897


core/string_utils.cpp
```diff
@@ -60,6 +60,50 @@ UString jsonnet_string_escape(const UString &str, bool single)
...
+        }
+        codepoint *= 16;
+        codepoint += digit;
+    }
+    return codepoint;
+}
+
+bool is_bmp_codepoint(const unsigned long codepoint)
+{
+    return codepoint != 0xd83d;
```
sbarzowski, line 93: This is insufficient. E.g. 0xd834 is also outside bmp. IIUC 0x0000 to 0xD7FF and 0xE000 to 0xFFFF are BMP characters and everything else is non-BMP.

test_suite/unicode_bmp1.json
```diff
@@ -0,0 +1 @@
+["\uD83D\uDCD8"]
```
sbarzowski, line 1: 1) It should be `unicode_bmp1.jsonnet`, not `.json`. It's parsed and treated as Jsonnet, so `.json` extension is a bit misleading here.
2) We'll need more examples like that which exercise our BMP-detection.

sbarzowski:
Thanks! This seems pretty close to what we need. Would you mind signing the CLA, so we can use this?

core/string_utils.cpp
rhowe: Yes, I was really just raising this as a hacky demo of the kinds of changes I thought might be needed. The way I split out the \u escape parsing is pretty vile.

rhowe:
@googlebot I signed it!

rhowe:
OK, what was a mere demo of the bug is now I think an actual functional implementation.

However there are issues which would make me wary about it being merged in its present state:

* I do not consider myself adept at C++ - far from it. Please review the code thoroughly and don't hesitate to rewrite it.
* The hacks I did to the parser to deal with the `\u` escapes are pretty horrible and I think someone more familiar with the parser should take a look and rework them.
* The test cases are obviously weak.

PR #902, opened 2021-03-26
Fix an error in stdlib doc `std.splitLimit` function

IanChen83:


IanChen83:
@googlebot I signed it!

sbarzowski:
Thanks!

Issue #868, opened 2020-12-02
"std.std.clamp" in docs


sbarzowski:
Remove duplicated `std.`.

PR #907, opened 2021-04-21
Fix misprint in stdlib ref

roidelapluie:
Signed-off-by: Julien Pivotto <roidelapluie@inuits.eu>


sbarzowski:
Thank you!  Fixes #868

PR #918, opened 2021-06-14
Fix compilation warning signed/unsigned

sthaha:
Fixes the follow compilation warning
```
core/vm.cpp:1614:31: warning: comparison of integer expressions of different signedness:
‘int’ and ‘std::vector<std::__cxx11::basic_string<char> >::size_type’
{aka ‘long unsigned int’} [-Wsign-compare]
```

```
 1614 :             for (int i = 0; i < v.size(); ++i) {
                         ^^^^^
```

Signed-off-by: Sunil Thaha <sthaha@redhat.com>

### DIFF
--- a/core/vm.cpp
+++ b/core/vm.cpp
@@ -343,7 +343,7 @@ class Stack {
      */
     virtual void dump(void)
     {
-        for (unsigned i = 0; i < stack.size(); ++i) {
+        for (std::size_t i = 0; i < stack.size(); ++i) {
             std::cout << "stack[" << i << "] = " << stack[i].location << " (" << stack[i].kind
                       << ")" << std::endl;
         }
@@ -970,7 +970,7 @@ class Interpreter {
                              const std::vector<Value> &args, const std::vector<Value::Type> params)
     {
         if (args.size() == params.size()) {
-            for (unsigned i = 0; i < args.size(); ++i) {
+            for (std::size_t i = 0; i < args.size(); ++i) {
                 if (args[i].t != params[i])
                     goto bad;
             }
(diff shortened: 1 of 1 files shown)


core/vm.cpp
```diff
@@ -1611,7 +1611,7 @@ class Interpreter {
 
             // Convert yaml to json and push onto json array
             ryml::Tree doc;
-            for (int i = 0; i < v.size(); ++i) {
+            for (unsigned i = 0; i < v.size(); ++i) {
```
sbarzowski, line 1614: Probably it's better to use `size_t` instead of `unsigned`.

sthaha: I am happy to make the change but elsewhere in the file, it uses `unsigned`  and I felt cleaning up to use `std::size_t` should be separate commit. 

```
 346          for (unsigned i = 0; i < stack.size(); ++i) {
973           for (unsigned i = 0; i < args.size(); ++i) {
2217         for (unsigned i = 0; i < ast.args.size(); ++i) {
...
```
I have changed all such usage to `std:size_t`

sthaha: Tested on 
```
❯ gcc --version
gcc (GCC) 10.3.1 20210422 (Red Hat 10.3.1-1)
Copyright (C) 2020 Free Software Foundation, Inc.
This is free software; see the source for copying conditions.  There is NO
warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
```
And `make test` works fine.

sbarzowski: Thanks for caring about consistency! Well, currently we have a mix of `unsigned` and `size_t` in this context. I'll be happy with a commit which changes all such usage to `size_t`.

sthaha: I have changed, just this file, I am sure there will be more. Perhaps in a different commit. My C++ is quite rusty now :D

sbarzowski:
Thanks!

PR #920, opened 2021-06-18
Add missing #!/bin/bash

jcpunk:
A handful of platforms don't like the missing `#!/bin/bash` when running the test script.

### DIFF
--- a/tests.sh
+++ b/tests.sh
@@ -1,3 +1,4 @@
+#!/usr/bin/env bash
 set -e
 
 JSONNET_BIN="${JSONNET_BIN:-./jsonnet}"
--- a/benchmarks/regen_benchmarks.sh
+++ b/benchmarks/regen_benchmarks.sh
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
 
 set -e
 set -x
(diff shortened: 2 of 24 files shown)


tests.sh
```diff
@@ -1,3 +1,4 @@
+#/bin/bash -x
```
sbarzowski, line 1: Good point, but I wouldn't use `-x` here. This script already displays messages about its status.

jcpunk: Pushed new version without -x

sbarzowski (approved):

sbarzowski:
Also while we're at it, I've found out that the proper way to do it is through `/usr/bin/env bash` as some more exotic system (e.g. NixOS, some BSDs) may not have `/bin/bash`.

jcpunk:
Hello,

I've reached out to my supervisor on getting the CLA signed.  It might take a bit....

jcpunk:
@googlebot I signed it!

jcpunk:
@googlebot I signed it!

jcpunk:
:) I think I'm all sorted now.

sbarzowski:
Awesome, thanks.

PR #921, opened 2021-06-22
Fix k8s test

timp87:
Make pytest a bit happier

### DIFF
--- a/case_studies/kubernetes/test_same.py
+++ b/case_studies/kubernetes/test_same.py
@@ -50,10 +50,10 @@ def canonicalize(doc):
 
 for filename in files:
     with open(filename + '.old.yaml', 'r') as f:
-        yaml_doc = canonicalize(yaml.load(f))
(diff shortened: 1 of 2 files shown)


timp87:
@googlebot I signed it!

case_studies/kubernetes/test_same.py
```diff
@@ -50,10 +54,10 @@ def canonicalize(doc):
 
 for filename in files:
     with open(filename + '.old.yaml', 'r') as f:
-        yaml_doc = canonicalize(yaml.load(f))
+        yaml_doc = canonicalize(yaml.load(f, Loader=yaml.FullLoader))
```
sbarzowski, line 57: I think it's better to use SafeLoader unless we have a good reason not to (in this case we're working with trusted data, so it shouldn't really matter, but let's stick with a better default).

timp87: Sure, done!

PR #923, opened 2021-06-22
CMake support for python extension

### LINKED ISSUE #922: CMake support for python extension
When building up the jsonnet libraries, it would be nice if I could also pass something like `BUILD_PYTHON3_SUPPORT:BOOL=ON` to get the python extension built at the same time.

And ideally, it would respect the `BUILD_SHARED_BINARIES:BOOL`....

jcpunk:
Fixes #922

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -11,6 +11,7 @@ option(BUILD_JSONNETFMT "Build jsonnetfmt command-line tool." ON)
 option(BUILD_TESTS "Build and run jsonnet tests." ON)
 option(BUILD_STATIC_LIBS "Build a static libjsonnet." ON)
 option(BUILD_SHARED_BINARIES "Link binaries to the shared libjsonnet instead of the static one." OFF)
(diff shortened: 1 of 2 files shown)


sbarzowski:
I'm not sure I understand what's going on here. Could you explain (either here or in the comments)?

jcpunk:
This adds the option to build the python extension as a part of the cmake process

For example with this patch:

```
mkdir build
cmake -DBUILD_PYTHON3_SUPPORT:BOOL=ON ..
make
make install
```

Would also result in the creation of the `_jsonnet` python library.  Similarly, if you set `-DBUILD_SHARED_BINARIES:BOOL=ON` too the python library would be built, but linked to `libjsonnet` (rather than have it statically complied in).

I'm in the process of building a package of jsonnet and would prefer to have a single build process to get all my objects (and have those objects link against `libjsonnet` rather than have it statically built in).

jcpunk:
I'm over thinking this.... I can just tweak the BUILD files as needed.  Ignore this....

PR #924, opened 2021-06-23
Add objectGetOrDefault and objectGetOrDefaultAll to std

ershov:


stdlib/std.jsonnet
```diff
@@ -1512,6 +1512,12 @@ limitations under the License.
   objectValuesAll(o)::
     [o[k] for k in std.objectFieldsAll(o)],
 
+  objectGetOrDefault(o, f, default = null)::
```
raphendyr, line 1515: Should these be in alphabetical order? I.e., after fields and before has.

test_suite/stdlib.jsonnet
```diff
@@ -168,6 +168,19 @@ std.assertEqual(std.objectValues({ x::: 1 } { x: 1 }), [1]) &&
...
+std.assertEqual(std.objectGetOrDefault({ x: 1, y: 2 }, 'z'), null) &&
+std.assertEqual(std.objectGetOrDefault({ x::: 1, y::: 2 }, 'x', 5), 1) &&
+std.assertEqual(std.objectGetOrDefault({ x::: 1, y::: 2 }, 'z', 5), 5) &&
+std.assertEqual(std.objectGetOrDefault({ x::: 1, y::: 2 }, 'z'), null) &&
+std.assertEqual(std.objectGetOrDefault({ x:: 1, y:: 2 }, 'x', 5), 5) &&
+std.assertEqual(std.objectGetOrDefault({ x:: 1, y:: 2 }, 'x'), null) &&
+std.assertEqual(std.objectGetOrDefault({ x:: 1, y:: 2 }, 'z', 5), 5) &&
+std.assertEqual(std.objectGetOrDefault({ x:: 1, y:: 2 }, 'z'), null) &&
+std.assertEqual(std.objectGetOrDefault({}, 'z', 5), 5) &&
+std.assertEqual(std.objectGetOrDefault({}, 'z'), null) &&
```
raphendyr, line 182: There are no tests for `objectGetOrDefaultAll`, should there be?

doc/ref/stdlib.html
```diff
@@ -401,7 +455,7 @@ <h3 id="math">
   <div class="hgroup-inline">
     <div class="panel">
       <h4 id="std.clamp">
-        std.clamp(x, minVal, maxVal)
+        std.std.clamp(x, minVal, maxVal)
```
raphendyr, line 458: Does this happen to be included by mistake?

doc/ref/stdlib.html
```diff
@@ -1146,6 +1200,39 @@ <h4 id="parseJson">
   </div>
 </div>
 
+<div class="hgroup">
+  <div class="hgroup-inline">
+    <div class="panel">
+      <h4 id="parseYaml">
+        std.parseYaml(str)
```
raphendyr, line 1207: Does this happen to be included by mistake?

raphendyr (commented):
Hi, I'm a random person from the internet, but I reviewed the PR as it's something I ended up writing myself for a project. Hence, I'm motivated of seeing this integrated in the stdlib!

Just few things I think you included by mistake. Good work! I hope it will get merged.

sbarzowski:
Many thanks for the PR and the 3rd party review!

I have two things I'm not sure about.
1) Do we need a variant which ignores fields which are not visible? Visibility is about how the object is serialized. When you access a specific field you normally don't care if it's visible or not.
2) `objectGetOrDefault` doesn't sound right to me – grammar-wise. I know that you're trying to follow the existing convention, but it "object fields` or "object values" or "object has field" are things that could appear in a normal sentence. "Object get some field" sounds weird (making an adverb out of "object...")... I know I'm rambling a bit, but I hope you get the idea. Maybe we can just have getOrDefault which does the same thing as `objectGetOrDefaultAll`? What do you think?

raphendyr:
> I know I'm rambling a bit, but I hope you get the idea. Maybe we can just have getOrDefault which does the same thing asobjectGetOrDefaultAll`? What do you think?

For the conversation, my own implementation uses name `getField` for this purpose.

ershov:
Thank you guys for your feedback.

@raphendyr The unrelated changes to HTML are by mistake on the one hand, on the other hand someone who had made changes to the source of the doc, didn't run the html generation script, hence there is a discrepancy building up. Not sure if that's done on purpose in order to delay the official doc changes or my mistake. Given that ".std" duplication typo, I'm leaning towards someone's obliviousness.

Also, it's not clear what version do I use as `availableSince`. The current version is not being incremented, it's kinda available in it but not in the versions compiled from the previous source. I would increment the minor version but I don't see it being done in this project. That's not good since there are quite a few versions with exactly same version number but different functionality. I reckon every commit with changes in the code must increment the (minor) version number.
So I'm just using the supposedly next version - hope it's ok.

By the way, `parseYaml` says it's available since "x.y.z" :)

I will remove these unrelated changes from this PR.

@sbarzowski With naming I'm just trying to be consistent and responsible. Once it's released, it can't be changed.

Currently, the naming scheme is such that functions for objects are prefixed by "object". It's longish but it makes sense to me: there are also functions for sets, manifests, etc. I think, if I don't start the function from "object", it probably should be more universal and be able to accept other types too. But right now it's specifically for objects, hence the prefix.

Next - I honestly find "All" suffix rather confusing but it's short (which is good) and it's what we have right now, so - no suffix for visible fields, "All" for all fields.
Whether or not the user will always want to consider all fields, I wouldn't be making such an assumption too quickly. That depends on why the user wants to use this function. For example, if they are constructing an external representation of the object, they will likely want to consider only visible fields.
I wouldn't want to both break consistency and reduce potentially useful functionality in one shot.

`getField`-alike name looks good but it looks similar to `getFields` which returns the field names whereas this function returns the field value. Not quite sure how confusing it might get.

But if we are feeling adventurous and up to change some trends, we could use just `get`.
Also, following the internal implementation of other "All"/"no-All" functions, might add a boolean argument which then gets passed straight to `std.objectHasEx`.

So it would be just one function:
```
  get(o, f, default = null, all = true)::
    if std.objectHasEx(o, f, all) then o[f] else default,
```

What do you guys think?

raphendyr:
> `get(o, f, default = null, all = true)::`

I personally like this. However, I would name `all` as `inc_hidden` (name used in `objectHasEx`), `incHidden` (if that is the preferred naming style) or just `hidden` for short.

Second best in my option could be `objectGet(..)`, as it would indicate that it's not about arrays.

sbarzowski:
> availableSince

The way to go is to put "upcoming" as the version number. When I do the release, I replace that with the current version number.


> > `get(o, f, default = null, all = true)::`
> 
> I personally like this. However, I would name `all` as `inc_hidden` (name used in `objectHasEx`), `incHidden` (if that is the preferred naming style) or just `hidden` for short.
> 
> Second best in my option could be `objectGet(..)`, as it would indicate that it's not about arrays.

This sounds very reasonable. I like `inc_hidden` more than `all`.

sbarzowski:
<digression>

> With naming I'm just trying to be consistent and responsible. Once it's released, it can't be changed.

Yes, thank you for being thoughtful. This is why we're keeping some names which are admittedly far from ideal. So we now have to do a little balancing act between convenience, consistency and explicitness.

One day, I imagine we can have stdlib 2.0, which we could introduce without breaking backwards compatibility. The user would need to start their program with something like `local std = std.std2` to use the new version (new standard library would be a sub-field of std). This is still slightly disruptive and might be confusing. 

If we do this, we'll get a chance to rethink our naming convention.

Generally I'm prioritizing work which doesn't disrupt the user in any way, so it's not going to happen very soon probably, unless someone volunteers.

</digression>

ershov:
@sbarzowski so what do we do next to this pull request?

sbarzowski:
@ershov Let's change to the version with one `get` function. Otherwise looks good.

sbarzowski:
Thanks a lot!

doc/ref/stdlib.html
seh: @ershov, did you correct this in a different PR?

ershov:
Cheers!

doc/ref/stdlib.html
ershov: @seh no I've included it in this one

seh: Where is it included? When I look at the diff against the parent branch, it still shows this PR bringing in an extra ".std" on line 458.

PR #926, opened 2021-06-28
Fix compiler warns

timp87:
Got 2 warnings in clang 11.

### DIFF
--- a/core/vm.cpp
+++ b/core/vm.cpp
@@ -1705,8 +1705,9 @@ class Interpreter {
                 abort();
             }
 
-            case json::value_t::binary:
-                break;
+            case json::value_t::binary: {
+                abort();
+            }
         }
     }
 
--- a/core/vm.cpp
+++ b/core/vm.cpp
@@ -1704,6 +1704,9 @@ class Interpreter {
             case json::value_t::discarded: {
                 abort();
             }
+
+            case json::value_t::binary:
+                break;
         }
     }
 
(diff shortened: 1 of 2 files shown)


timp87:
Ah, I used external nlohmann-json of version 3.9.1

core/vm.cpp
```diff
@@ -1704,6 +1704,9 @@ class Interpreter {
             case json::value_t::discarded: {
                 abort();
             }
+
+            case json::value_t::binary:
```
sbarzowski, line 1708: This should never happen (as this is a non-standard extension which doesn't occur when parsing JSON normally). I don't think just breaking in such case is reasonable. Let's have `abort()` here.

core/vm.cpp
```diff
@@ -1704,6 +1704,10 @@ class Interpreter {
             case json::value_t::discarded: {
                 abort();
             }
+
+            case json::value_t::binary: {
```
hassamchohan-smava, line 1708: @timp87 Hello, This change here has some breaking changes on our CI/CD pipelines since few hours ago. I will paste the error snippet down below. Could you please fix this at your earliest convenience ? 

```core/vm.cpp: In member function 'void {anonymous}::Interpreter::otherJsonToHeap(const json&, bool&, {anonymous}::Value&)':
core/vm.cpp:1714:18: error: 'binary' is not a member of 'nlohmann::basic_json<>::value_t {aka nlohmann::detail::value_t}'
             case json::value_t::binary: {
                  ^
Makefile:159: recipe for target 'core/vm.o' failed
make: *** [core/vm.o] Error 1

sparkprime: I reverted it, thanks for the heads up

hassamchohan-smava: Thanks for taking care of it.

ptimofee: I think the problem is the same as stated in my last comment.
Seems like you're bundled nlohmann-json is too old.

PR #929, opened 2021-06-30
Add tailstrict to std *stripChars functions

raphendyr:
Current implementation will fail with `max stack frames exceeded` for provided simple test cases. This PR fixes that by adding the tail recursion flag. I think this was just missing.

Sadly, testing the stack limit is very time consuming. I think it could be ok to drop the provided tests too. Times on my machine (`time make test` after `make`):

 * `make test  39.70s user 5.38s system 99% cpu 45.507 total` for original tests
 * `make test  52.49s user 5.02s system 99% cpu 57.874 total` for only `rstripChars`
 * `make test  81.86s user 5.35s system 99% cpu 1:27.63 total` for both (the provided commit)

### DIFF
--- a/benchmarks/bench.09.jsonnet
+++ b/benchmarks/bench.09.jsonnet
@@ -1,0 +1,8 @@
+// This string must be longer than max stack frames
+local veryLongString = std.join('', std.repeat(['e'], 510));
+
+std.assertEqual(std.stripChars(veryLongString + 'ok' + veryLongString, 'e'), 'ok') &&
+std.assertEqual(std.lstripChars(veryLongString + 'ok', 'e'), 'ok') &&
+std.assertEqual(std.rstripChars('ok' + veryLongString, 'e'), 'ok') &&
+
+true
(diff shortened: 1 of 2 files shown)


raphendyr:
@googlebot I signed it!

test_suite/stdlib.jsonnet
```diff
@@ -203,20 +203,25 @@ std.assertEqual(std.endsWith('food', 'food'), true) &&
 std.assertEqual(std.endsWith('food', 'omgfood'), false) &&
 std.assertEqual(std.endsWith('food', 'wat'), false) &&
 
+local veryLongString = std.join('', std.repeat(['e'], 510));
```
sbarzowski, line 206: This is too slow for the regular tests. Please add it to `benchmarks/` instead (as a separate file). 

(BTW this really shouldn't be so damn slow. I wonder how go-jsonnet fares here).

raphendyr: I moved the test to benchmars.

PR #934, opened 2021-07-26
Fix double definition of libname 'std.std.clamp/3'

bryanhuntesl:
26th July 2021
https://jsonnet.org/ref/stdlib.html reads: 

```
std.std.clamp(x, minVal, maxVal)
Available since version 0.15.0.
```

Should read: 
```
std.clamp(x, minVal, maxVal)
Available since version 0.15.0.
```


bryanhuntesl:
@googlebot I signed it!

sbarzowski:
Thank you!

PR #939, opened 2021-09-02
Add official OSI name in the license metadata

ecederstrand:
This makes it easier for automatic license checkers to verify the license of this package.


sbarzowski:
Thanks!

PR #940, opened 2021-09-04
use WASM go-jsonnet build on the website

sparkprime:
use WASM go-jsonnet build on the website


sbarzowski:
LG, but this change doesn't actually include WASM libjsonnet itself. I think we should add it in the same commit (or if we want to keep it separate, add it before), so that the website stays deployable.

sparkprime:
@sbarzowski I committed the Golang one, this one has a dubious workaround for jekyll so can you take a look? Thanks

sbarzowski (approved):

sbarzowski:
Looks good. The Jekyll workaround isn't too bad. We could have a script ready for starting this server in one easy command.

PR #955, opened 2021-12-20
Remove extra 'std.' prefix from docs for std.clamp

rhowe:


rhowe:
No double standards here!

PR #961, opened 2021-12-29
Add all/any implemented using foldl.

### LINKED ISSUE #954: Add std.all(arr) and std.any(arr)
I'm a bit surprised there's no short hand for testing a list of boolean values in Jsonnet.

I'll admit it can be trivially implemented using `std.foldl` but still it would be convenient if it can be implemtend as part of the stdlib:

```jsonnet
// all/any implemented using std.foldl
local all(arr) = std.foldl(function(x, y) x && y, arr, true);
local any(arr) = std.foldl(function(x, y) x || y, arr, false);

{
  test_all: all([true, false]),
  test_any: any([true, false]),
}

/*
$ jsonnet test.jsonnet 
{
   "test_all": false,
   "test_any": true
}
*/
```

Example usecase: construct a predicate with multiple rules:

```jsonnet
local predicate(x) = all([
  x.foo == 'a',
  x.bar == 'b',
])
```

yegle:
Fixes #954

### DIFF
--- a/doc/_stdlib_gen/stdlib-content.jsonnet
+++ b/doc/_stdlib_gen/stdlib-content.jsonnet
@@ -1244,6 +1244,24 @@ local html = import 'html.libsonnet';
             |||,
           ]),
         },
+        {
+          name: 'all',
+          params: ['arr'],
+          description: html.paragraphs([
+            |||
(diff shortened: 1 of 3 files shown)


sbarzowski:
Nice. Thanks!

Issue #965, opened 2021-12-31
Problematic reference in std lib. Different results from same code.


KetchupBomb:
Hello Jsonnet,

I am preparing to submit a PR for expanding `std.splitLimit()` (and by extensions `std.split()`) to support multi-character delimiters.

<details>
  <summary>Example</summary>

```shell
# Current behavior
$ echo 'std.split("foo//bar", "//")' | jsonnet -
RUNTIME ERROR: std.split second parameter should have length 1, got 2
        std.jsonnet:(115:5)-(116:31)    function <anonymous>
        <stdin>:1:1-28

# PR behavior
$ echo 'std.split("foo//bar", "//")' | jsonnet -
  - foo
  - bar
```
</details>

However, I am getting inconsistent results. If I take my source code, and put in a file and run by itself, it works. But if I take the same code and slap it in to `stdlib/std.jsonnet`, it doesn't work.

<details>
  <summary>Working, local development</summary>

```json
test:
  - foo
  - bar
```

  <details>
    <summary>Source file</summary>

```jsonnet
{
  split(str, c)::
    assert std.isString(str) : 'std.split first parameter should be a string, got ' + std.type(str);
    assert std.isString(c) : 'std.split second parameter should be a string, got ' + std.type(c);
    $.splitLimit(str, c, -1),

  splitLimit(str, c, maxsplits)::
    assert std.isString(str) : 'std.splitLimit first parameter should be a string, got ' + std.type(str);
    assert std.isString(c) : 'std.splitLimit second parameter should be a string, got ' + std.type(c);
    assert std.isNumber(maxsplits) : 'std.splitLimit third parameter should be a number, got ' + std.type(maxsplits);
    local aux(str, delim, i, arr, v) = 
      local c = str[i : i + std.length(delim) : 1];
      if i >= std.length(str) then
        arr + [v]
      else if c == delim && (maxsplits == -1 || std.length(arr) < maxsplits) then
        aux(str, delim, i + std.length(c), arr + [v], '') tailstrict
      else
        aux(str, delim, i + 1, arr, v + str[i]) tailstrict;

    aux(str, c, 0, [], ''),

  test: $.split('foo//bar', '//'),
}
```
  </details>
</details>

<details>
  <summary>Non-working fork</summary>

```shell
$ git branch -v
* master 90a9f9a Explicitly define what the super keyword means in tutorial (#960)

$ make clean && make

$ echo 'std.split("foo//bar", "//")' | ./jsonnet -
[
[...]

PR #967, opened 2022-01-02
split and splitLimit to support multi-character strings

KetchupBomb:
Per https://github.com/google/jsonnet/issues/965, adds the ability to split on multi-character strings. Just a few notes:

- Updated both Jsonnet and underlying C++ code.
- ~~Changed `c` parameter name to `pat` to better match existing semantics of `std.findSubstr()`.~~
- Changed `aux()` signature to only pass required parameters, referring to all others from outside the `local` scope.

### DIFF
--- a/doc/_stdlib_gen/stdlib-content.jsonnet
+++ b/doc/_stdlib_gen/stdlib-content.jsonnet
@@ -339,36 +339,46 @@ local html = import 'html.libsonnet';
         {
           name: 'split',
           params: ['str', 'c'],
-          description: |||
-            Split the string <code>str</code> into an array of strings, divided by the single character
-            <code>c</code>.
-          |||,
+          description: [
+            html.p({}, |||
+              Split the string <code>str</code> into an array of strings, divided by the string
+              <code>c</code>.
+            |||),
+            html.p({}, |||
+              Note: Versions up to and including 0.18.0 require <code>c</code> to be a single character.
+            |||),
+          ],
           examples: [
             {
-              input: @'std.split("foo/bar", "/")',
-              output: std.split('foo/bar', '/'),
+              input: @'std.split("foo/_bar", "/_")',
+              output: std.split('foo/_bar', '/_'),
             },
             {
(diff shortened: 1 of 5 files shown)


doc/_stdlib_gen/stdlib-content.jsonnet
```diff
@@ -338,37 +338,37 @@ local html = import 'html.libsonnet';
         },
         {
           name: 'split',
-          params: ['str', 'c'],
+          params: ['str', 'pat'],
```
sbarzowski, line 341: 1) IMO `sep` (separator) would be a better name than `pat`.
2) Technically changing the argument name is a breaking change. In case I don't think it's very reasonable to use it with a named argument, so it's fine. We need to be careful with that.

doc/_stdlib_gen/stdlib-content.jsonnet
```diff
@@ -338,37 +338,37 @@ local html = import 'html.libsonnet';
...
           description: |||
-            Split the string <code>str</code> into an array of strings, divided by the single character
-            <code>c</code>.
+            Split the string <code>str</code> into an array of strings, divided by the string
+            <code>pat</code>.
```
sbarzowski, line 344: We should note that up to and including 0.18.0 require the separator to be single character.

sbarzowski (approved):
LG modulo (minor) comments.

doc/_stdlib_gen/stdlib-content.jsonnet
KetchupBomb: I'm fine to leave as `c` if you think that's the right call. I only was thinking about how `std.findSubstr()` used "`pat`" rather than "`c`" and wanted to make it clearer that multi-character strings are supported. I hadn't even considered those who use named parameters...

Probably best to leave as `c` then, for sheer compatibility sake?

sbarzowski: Yeah, let's leave as `c`.

sbarzowski:
Thank you!

PR #972, opened 2022-01-08
splitLimitR in standard library

KetchupBomb:
See title. Split strings moving from right to left.

### DIFF
--- a/doc/_stdlib_gen/stdlib-content.jsonnet
+++ b/doc/_stdlib_gen/stdlib-content.jsonnet
@@ -383,6 +383,18 @@ local html = import 'html.libsonnet';
           ],
         },
         {
+          name: 'splitLimitR',
+          params: ['str', 'c', 'maxsplits'],
+          availableSince: 'upcoming',
+          description: 'As <code>std.splitLimit(str, c, maxsplits)</code> but will split from right to left.',
+          examples: [
+            {
+              input: @'std.splitLimitR("/_foo/_bar", "/_", 1)',
+              output: std.splitLimitR('/_foo/_bar', '/_', 1),
+            },
+          ],
+        },
+        {
           name: 'strReplace',
           params: ['str', 'from', 'to'],
           description: |||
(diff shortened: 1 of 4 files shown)


doc/_stdlib_gen/stdlib-content.jsonnet
```diff
@@ -382,6 +382,24 @@ local html = import 'html.libsonnet';
...
         },
+        {
+          name: 'splitLimitR',
+          params: ['str', 'c', 'maxsplits'],
+          description: [
+            html.p({}, |||
+              As <code>std.splitLimit(str, c, maxsplits)</code> but will split from right to left.
+            |||),
+            html.p({}, |||
+              Note: Versions up to and including 0.18.0 require <code>c</code> to be a single character.
```
sbarzowski, line 393: This doesn't seem relevant as it's a new function. Instead, there should be 'available_since' field present.

stdlib/std.jsonnet
```diff
@@ -132,6 +132,18 @@ limitations under the License.
...
+  splitLimitR(str, c, maxsplits)::
+    assert std.isString(str) : 'str.splitLimitR first parameter must be a String, got ' + std.type(str);
+    assert std.isString(c) : 'str.splitLimitR second parameter must be a String, got ' + std.type(c);
+    assert std.length(c) >= 1 : 'std.splitLimitR second parameter must have length 1 or greater, got ' + std.length(c);
+    assert std.isNumber(maxsplits) : 'str.splitLimitR third parameter must be a Number, got ' + std.type(maxsplits);
+    if maxsplits == -1 then
+      std.splitLimit(str, c, -1)
+    else
+      local revStr(str) = std.join('', std.reverse(str));
+      local revArr(arr) = std.reverse(arr);
```
sbarzowski, line 144: nit: no need for this function. Just use `std.reverse` directly.

sbarzowski:
Thanks, that's a good addition. Other than the comments, looks good.

PR #974, opened 2022-01-13
Any all optimization

KetchupBomb:
`std.any()` and `std.all()` iterate over all items in their input when it can be short circuited. This has large speed gains when working with large lists:

<details>
  <summary><code>std.any()</code></summary>

```shell
# Before code change
$ time echo 'std.any(std.map(function(x) false, std.range(0, 100000)) + [true])' | bazel run -c opt //cmd:jsonnet 2> /dev/null -- -
true

real    0m3.775s
user    0m3.626s
sys     0m0.070s

$ time echo 'std.any([true] + std.map(function(x) false, std.range(0, 100000)))' | bazel run -c opt //cmd:jsonnet 2> /dev/null -- -
true

real    0m3.820s
user    0m3.676s
sys     0m0.061s


# After code change
$ time echo 'std.any(std.map(function(x) false, std.range(0, 100000)) + [true])' | bazel run -c opt //cmd:jsonnet 2> /dev/null -- -
true

real    0m3.306s
user    0m3.196s
sys     0m0.031s

$ time echo 'std.any([true] + std.map(function(x) false, std.range(0, 100000)))' | bazel run -c opt //cmd:jsonnet 2> /dev/null -- -
true

real    0m0.190s
user    0m0.099s
sys     0m0.011s
```
</details>

<details>
  <summary><code>std.all()</code></summary>

```shell
# Before code change
$ time echo 'std.all(std.map(function(x) true, std.range(0, 100000)) + [false])' | bazel run -c opt //cmd:jsonnet 2> /dev/null -- -
false

real    0m3.865s
user    0m3.757s
sys     0m0.031s

$ time echo 'std.all([false] + std.map(function(x) true, std.range(0, 100000)))' | bazel run -c opt //cmd:jsonnet 2> /dev/null -- -
false

real    0m3.896s
user    0m3.765s
sys     0m0.060s


# After code change
$ time echo 'std.all([false] + std.map(function(x) true, std.range(0, 100000)))' | bazel run -c opt //cmd:jsonnet 2> /dev/null -- -
false

real    0m0.180s
user    0m0.102s
sys     0m0.010s

$ time echo 'std.all(std.map(function(x) true, std.range(0, 100000)) + [false])' | bazel run -c opt //cmd:jsonnet 2> /dev/null -- -
false

real    0m3.408s
user    0m3.280s
sys     0m0.051s
```
</details>

### DIFF
--- a/stdlib/std.jsonnet
+++ b/stdlib/std.jsonnet
@@ -1608,11 +1608,18 @@ limitations under the License.
     if !std.isArray(arr) then
       error 'any() parameter should be an array, got ' + std.type(arr)
     else
-      local or = function(x, y)
-        if !std.isBoolean(y) then
-          error std.format('element "%s" of type %s is not a boolean', y, std.type(y))
-        else x || y;
-      std.foldl(or, arr, false),
+      local arrLen = std.length(arr);
+      local aux(idx) =
+        local e = arr[idx];
+        if idx >= arrLen then
+          false
+        else if !std.isBoolean(e) then
+          error std.format('element "%s" of type %s is not a boolean', e, std.type(e))
+        else if e then
+          true
+        else
+          aux(idx + 1) tailstrict;
+      aux(0),
 
   // Three way comparison.
   // TODO(sbarzowski): consider exposing and documenting it properly
--- a/stdlib/std.jsonnet
+++ b/stdlib/std.jsonnet
@@ -1599,11 +1599,19 @@ limitations under the License.
     if !std.isArray(arr) then
       error 'all() parameter should be an array, got ' + std.type(arr)
     else
-      local and = function(x, y)
-        if !std.isBoolean(y) then
-          error std.format('element "%s" of type %s is not a boolean', y, std.type(y))
-        else x && y;
-      std.foldl(and, arr, true),
+      local arrLen = std.length(arr);
+      local aux(idx) =
+        local e = arr[idx];
+        if idx >= arrLen then
+          true
+        else if !std.isBoolean(e) then
+          error std.format('element "%s" of type %s is not a boolean', e, std.type(e))
+        else if !e then
+          false
+        else
+          aux(idx + 1) tailstrict;
+      aux(0),
+
   any(arr)::
     if !std.isArray(arr) then
       error 'any() parameter should be an array, got ' + std.type(arr)
--- a/stdlib/std.jsonnet
+++ b/stdlib/std.jsonnet
@@ -1601,15 +1601,16 @@ limitations under the License.
     else
       local arrLen = std.length(arr);
       local aux(idx) =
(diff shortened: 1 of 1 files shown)


stdlib/std.jsonnet
```diff
@@ -1599,20 +1599,35 @@ limitations under the License.
...
     else
-      local and = function(x, y)
-        if !std.isBoolean(y) then
-          error std.format('element "%s" of type %s is not a boolean', y, std.type(y))
-        else x && y;
-      std.foldl(and, arr, true),
+      local arrLen = std.length(arr);
+      local aux(idx) =
+        local e = arr[idx];
+        if idx >= arrLen then
```
seh, line 1605: What happens to that array index operation when the supplied index exceeds its length minus one?

stdlib/std.jsonnet
```diff
@@ -1599,20 +1599,35 @@ limitations under the License.
...
     else
-      local or = function(x, y)
-        if !std.isBoolean(y) then
-          error std.format('element "%s" of type %s is not a boolean', y, std.type(y))
-        else x || y;
-      std.foldl(or, arr, false),
+      local arrLen = std.length(arr);
+      local aux(idx) =
+        local e = arr[idx];
+        if idx >= arrLen then
```
seh, line 1622: Likewise, what happens to that array index operation when the supplied index exceeds its length minus one?

stdlib/std.jsonnet
KetchupBomb: While not obvious, and therefore could be addressed, this doesn't cause a runtime issue. Since Jsonnet is evaluated lazily, and since `e` isn't used until after the index check, it won't throw an exception:

```shell
$ echo 'std.any([])' | bazel run -c opt //cmd:jsonnet 2> /dev/null -- -
false

$ echo 'std.all([])' | bazel run -c opt //cmd:jsonnet 2> /dev/null -- -
true
```

seh: That was definitely not obvious to me, just reading the code. What led you to hoist that index operation up like that? Consider this slight revision:
```jsonnet
local aux(idx) =
  if idx >= arrLen then
    true
  else
    local e = arr[idx];
    if !std.isBoolean(e) then
      error std.format('element "%s" of type %s is not a boolean', e, std.type(e))
    else if !e then
      false
    else
      aux(idx + 1) tailstrict;
```

KetchupBomb: I don't mind changing to make it more clear. I will amend the PR later.

As for why I made the choice, I probably didn't notice the array index check. And I was further biased by the unit tests passing. 🙃

tachitworapong (approved):

sbarzowski:
Thanks to @KetchupBomb and reviewers!

PR #976, opened 2022-01-30
Add `importbin` statement

anguslees:
Add `importbin`.  Like `importstr` but returns an array of
numbers (integers 0-255).

Note! This changes the `JsonnetImportCallback` function type
signature, and requires modification to existing code that uses custom
import callbacks.  The included python bindings have been updated.


anguslees:
Reviewers: It's been a long time since I last wrote C++ - please treat everything with suspicion.

Re the import callback change:  We could add a new/additional binary-safe import callback, and fallback to the existing c-string import callback.  That will keep existing code compatibility, but I _think_ it will be worse overall, since we now have an `importbin` statement that will fail at runtime in some jsonnet environments without warning.   Happy to re-implement this PR if we want to follow some other migration strategy.

mackintosh-dev:
Question are you part of the development team involved with email address dcgetsit818@g mail. com 
Important if you could get right back to me

include/libjsonnet.h
```diff
@@ -69,10 +69,12 @@ void jsonnet_string_output(struct JsonnetVm *vm, int v);
...
  *     be resolved correctly.  Allocate memory with jsonnet_realloc.  Only use when *success = 1.
  * \param success Set this byref param to 1 to indicate success and 0 for failure.
- * \returns The content of the imported file, or an error message.
+ * \param buf Set this byref param to the content of the imported file, or an error message.  Allocate memory with jsonnet_realloc.  Do not include a null terminator byte.
+ * \param buflen Set this byref param to the length of the data returned in buf.
+ * \returns 0 to indicate success and 1 for failure.  On success, the content is in *buf.  On failure, an error message is in *buf.
  */
-typedef char *JsonnetImportCallback(void *ctx, const char *base, const char *rel, char **found_here,
-                                    int *success);
+typedef int JsonnetImportCallback(void *ctx, const char *base, const char *rel,
```
sbarzowski, line 76: I would deprecate old version and add a separate new one with a new function to register it.

Rationale: The main point of C++ implementation now is to avoid breaking existing users.

primeroz:
Hi, sorry for pinging like this on this PR 

I am using a version of `kubecfg` with `go-jsonnet` from master so that i can use importbin ( and it is working as expected ) but this is of course breaking any `jsonnet` and `jsonnetfmt` command i run since support for importbin is not merged in it yet.

Is there any ETA / plan to merge this ? 

thanks again

sparkprime:
What's the status exactly?  If it's in go-jsonnet already then you can use jsonnet and jsonnetfmt from there.

primeroz:
I did not know there was a `jsonnetfmt` from there i will try that

sparkprime:
Having said that, I am in favour of getting this PR in :)

PR #1011, opened 2022-07-30
Add ability to refer to subfields of objects via dot notation in string formatting

fedya-at-db:
You can now do something like this:
`"%(a.b)s" % {a:{b:"hello world"}}` which yields "hello world"

This is similar to python3's format strings behavior, though much simpler in capability. Every subfield must be an object except the last one.

Tested:
  added new happy path format test, and an error test


fedya-at-db:
@sparkprime Hi, wanted to check what you thought of this PR.

sparkprime:
I only just noticed this. It seems like it could be useful for a fair few cases but it's a bit too big for the release I'm just doing.

stdlib/std.jsonnet
```diff
@@ -768,11 +768,24 @@ limitations under the License.
...
+              local key = f[start:end];
+              if std.objectHasAll(obj, key) then
+                get_nested_val(obj[key], f, end+1, end+1) tailstrict
+              else
+                error 'No such partial field: ' + f[:end]
             else
-              error 'No such field: ' + f;
+              get_nested_val(obj, f, start, end+1) tailstrict
+          ;
+          local val = get_nested_val(obj, f);
```
CertainLach, line 788: Oh, this thing breaks some of my code, where I use formatting on object with '.' in field names, i.e
`"%(a.b)d" % {'a.b': 2}`
Shouldn't this be
`local val = if std.objectHasAll(obj, f) then obj[f] else get_nested_val(obj, f)`
?

fedya-at-db: That's probably better given jsonnet is allowing '.' in field names already but is a little hacky. Probably better than current code though to keep backwards compatibility. I was thinking of mimicking this after python's f'{a.b}' which wouldn't be interpreted as a single field so I just never considered that.

@sparkprime should probably decide which approach is more important but I think supporting the existing use case makes sense.

sparkprime: fortunate that you did not have a `)` in your field name I guess :)

I suppose the better fix is to have a way to put arbitary expressions inside strings, like you can do with javascript now.

Duologic: I've made an attempt at looking up values with JSONPath, it's not exactly the same as it moves the expression out of the string but should accomplish the same results. 

Have a look at it here: https://github.com/jsonnet-libs/xtd/blob/master/docs/jsonpath.md#fn-getjsonpath

PR #1020, opened 2022-10-06
Add #include <cstdint> for gcc-13 builds

xen0n:
See https://gcc.gnu.org/gcc-13/porting_to.html#header-dep-changes.

This fixes Gentoo bug https://bugs.gentoo.org/875569.


sbarzowski:
Thank you

PR #1039, opened 2022-11-29
Add RegEx support using RE2 (rebased #665)

Duologic:
(rebased #665)

---

Introduces 5 new built-in methods to the stdlib:
- `regexFullMatch(pattern, str)` -- Full match regex
- `regexPartialMatch(pattern, str)` -- Partial match regex
- `regexQuoteMeta(str)` -- Escape regex metachararacters
- `regexReplace(str, pattern, to)` -- Replace single occurance using regex
- `regexGlobalReplace(str, pattern, to)` -- Replace globally using regex

Since both `regexFullMatch` and `regexPartialMatch` can perform captures these functions return a "match" object upon match or `null` otherwise. For example:

```
$ ./jsonnet -e 'std.regexFullMatch("h(?P<mid>.*)o", "hello")'
{
   "captures": [
      "ell"
   ],
   "namedCaptures": {
      "mid": "ell"
   },
   "string": "hello"
}
```

Introduces a dependency on RE2 2019-06-01.
Builds tested using make, CMake and Bazel on Ubuntu 18.04.


Duologic:
Anything else I should address? @sparkprime @sbarzowski

Makefile
```diff
@@ -32,8 +32,7 @@ CXXFLAGS += -Iinclude -Ithird_party/md5 -Ithird_party/json -Ithird_party/rapidya
 CFLAGS ?= -g $(OPT) -Wall -Wextra -pedantic -std=c99 -fPIC
 CFLAGS += -Iinclude
 MAKEDEPENDFLAGS += -Iinclude -Ithird_party/md5 -Ithird_party/json -Ithird_party/rapidyaml/rapidyaml/src/ -Ithird_party/rapidyaml/rapidyaml/ext/c4core/src/
-LDFLAGS ?=
-
+LDFLAGS ?= -lre2
```
sbarzowski, line 35: This is the most controversial part. Introducing a dependency here may break existing users. The fix is easy for them, so maybe not a very big deal. Going from 0 to some dependencies might still be problematic in some contexts, but it's probably worth it for regexps.

We definitely need to add a README section for dependencies, though. It would also be good to check that the Python package works as expected and maybe mention the dependencies in the description.

@sparkprime What do you think?

test_suite/stdlib.jsonnet
```diff
@@ -1530,4 +1529,74 @@ std.assertEqual(std.all([true, false]), false) &&
...
+  {
+    string: 'hello',
+    captures: ['ll'],
+    namedCaptures: {
+      mid: 'll',
+    },
+  }
+) &&
+
+std.assertEqual(std.regexQuoteMeta(@'1.5-2.0?'), '1\\.5\\-2\\.0\\?') &&
```
sbarzowski, line 1594: This quotes a string to be included in regexp verbatim, right?

I'm not sure if we want to offer this. Is there a very compelling use case? I would prefer to err on the side of having smaller API (easier to add than to remove).

sbarzowski:
Sorry it took so long to reply. Looks mostly good. Left some minor comments.

sparkprime:
After I've done the current release (hopefully today) let's move forward on regex support in both impls for the next release.

sbarzowski:
> After I've done the current release (hopefully today) let's move forward on regex support in both impls for the next release.

Sounds good!

CertainLach (commented):
I think there should be an ability to process all matches, not only the first one, by returning matched indexes, or something like that?

Makefile
sparkprime: I'll look into that when I do the release.  I presume it will break existing windows build processes but whoever uses that ought to be able to figure it out.

test_suite/stdlib.jsonnet
sparkprime: Given that we are already locking into RE2, reducing the API surface isn't going to make it any easier to migrate to a different regex provider.  In terms of implementation, it's a small wrapper around the underlying call.

In terms of usefulness, if you want to search for a computed something inside something else I suppose it would be invaluable.

I'm ok with it.

core/vm.cpp
```diff
@@ -1440,6 +1450,129 @@ class Interpreter {
...
+        RE2 re(pattern, RE2::CannedOptions::Quiet);
+        if (!re.ok()) {
+            std::stringstream ss;
+            ss << "Invalid regex '" << re.pattern() << "': " << re.error();
+            throw makeError(stack.top().location, ss.str());
+        }
+
+        int num_groups = re.NumberOfCapturingGroups();
+
+        std::vector<std::string> rcaptures(num_groups);
```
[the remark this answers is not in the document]
sparkprime: If the underlying library returns an empty string then it's clearly "OK" to do that.  Would we choose to differ?  I'm not sure there's much point, because you have to do `!= null` instead of `!= ''` right?  So it doesn't help the programmer to use `null` instead.  On the other hand, it means there is a difference between the underlying library and the Jsonnet manifestation of it.  That difference would have to be documented by us and remembered by everyone else.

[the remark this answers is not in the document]
sparkprime: You could also make the argument that the only reason the RE2 C++ API returns the empty string is because it is much harder to return a sentinel in that language.  I wonder what the Go library does? @rohitjangid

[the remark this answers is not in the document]
CertainLach: Well, empty match is not the same as the optional group is not matched.
Howewer, looks like both C++ re2 and golang regex implementations don't care, and leave default type value for such matches, unlike Rust (https://docs.rs/regex/latest/regex/struct.Captures.html#method.get) and js regexps.

MrG9090 (approved):

PR #1040, opened 2022-12-01
Not a templating language

malcolmholmes:
The website describes Jsonnet as a "templating language". For a broader definition of "templating", this is correct. However, with templating tools such as Jinja or Helm around that treat structured data as flat text files, the term suggests that Jsonnet is less powerful than it is.

This suggestion arose from a conversation with a colleague who refused to use "another templating language". I'm suggesting this description harms Jsonnet, by limiting it.

Other possible descriptions:
* a purely functional language (with OO features)
* A powerful DSL for elegant description of JSON data
* A powerful data rendering language

### DIFF
--- a/doc/index.html
+++ b/doc/index.html
@@ -1,6 +1,6 @@
 ---
 layout: default
-title: The Data Templating Language
+title: A Functional Data Rendering Language
 ---
 
 <div class=hgroup>
@@ -11,7 +11,7 @@ title: The Data Templating Language
           <td><img id=big-logo src=img/isologo.svg></td>
           <td class=side-text>
             <p>
-              A data templating language<br>
+              A data rendering language<br>
               for app and tool developers
             </p>
             <ul>


Duologic (approved):
I like this, it's a bit more up-to-date with how the ecosystem around jsonnet evolved.

sparkprime:
I wonder if we should just use "configuration language" at this point.  The term "data templating language" dates back to 2014 when we were discussing how to describe to the outside world (i.e. not Google) what Jsonnet is for, because we felt that configuration language (which is what it was called inside Google) would not be easily understood externally.  But I feel that now, the term is much more widely used outside of Google.

malcolmholmes:
"configuration language" works for me, just not "templating language"!!

sparkprime:
Are you testing or validating configuration? :)

Duologic:
Not necessarily, it could validate JSON output from an API call piped into jsonnet.

My point is that 'configuration language' is still quite restricting for how powerful Jsonnet can be. Rendering configuration is one of the main features, by putting that in the initial tagline it might rule out any other uses prematurely, just like 'templating'.

sparkprime:
True, and I can see how a hermetic language would be a decent choice for testing / verification as well as configuration.  Neither of those are really "rendering" either, though.  Maybe we should single out configuration as the primary use case but also prominently draw attention to other use cases on the front page.

rhowe:
> since this is a config language

https://github.com/google/jsonnet/pull/607#issuecomment-467418904

:)

Issue #865, opened 2020-11-28
[feature request]: allow iterating over object in comprension (like dict comprehension in python)


bendavis78:
It would be useful to be able to do something like dict comprehensions in python, eg: `for key, value in obj`

A use-case would be when defining environment variables for a kubernetes container like so:
```jsonnet
{
  env:: {
    FOO: 'foo',
    BAR: 'bar',
    BAZ: 'baz'
  },
  containerEnv: [{name: k, value: v} for k, v in $.env]
}
```
->
```json
{"containerEnv": [
  {"name": "FOO", "value": "foo"},
  {"name": "BAR", "value": "bar"},
  {"name": "BAZ", "value": "baz"}
]}

sbarzowski:
You can do this currently with `std.objectFields`/`std.objectFieldsAll`. See also: #543.

Maybe we could add a shorthand syntax for that like in your example.

PR #1062, opened 2023-04-09
Add objectKeysValues, objectKeysValuesAll

zephyros-dev:
This PR adds two helper functions for extracting an array of objects with keys and values from an object. It's a simple function but I think it's worth putting into the stdlib given its appearance in numerous feature requests https://github.com/google/jsonnet/issues/865, https://github.com/google/jsonnet/issues/543

### DIFF
--- a/doc/_stdlib_gen/stdlib-content.jsonnet
+++ b/doc/_stdlib_gen/stdlib-content.jsonnet
@@ -98,6 +98,15 @@ local html = import 'html.libsonnet';
           |||,
         },
         {
+          name: 'objectKeysValues',
+          params: ['o'],
+          availableSince: 'upcoming',
+          description: |||
+            Returns an array of objects from the given object, each object having two fields: 
+            <code>key</code> (string) and <code>value</code> (object). Does not include hidden fields.
+          |||,
+        },
+        {
           name: 'objectHasAll',
           params: ['o', 'f'],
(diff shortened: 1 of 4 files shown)


KetchupBomb:
@zephyros-dev, this PR adds a bit of naming complexity to the std library. There are two existing functions:
- `std.objectFields()`
- `std.objectValues()`

Your addition of:
- `std.objectKeysValues()` 

introduces the term "keys" into the std library where the convention refers to them as "fields". 

It's minor, but I appreciate consistency. Do you have opinions on changing it to:
- `std.objectFieldsValues()`

zephyros-dev:
I also thought about naming it with objectFieldsValues when I first made the PR, but I changed my mind after looking through the feature requests #865 and #543. In there I see that the main terms used by both the maintainers and the commenters are keys and values, and personally I do prefer the terms since it is also used in other programming language like Javascript or Python.
I do think the naming should be more consistent, but changing the function names to objectFieldsValues means changing the signature of the return object to `{'field': 'foo', 'value': 'bar'}` if we want to be consistent (Right now it's `{'key': 'foo', 'value': 'bar'}` which map to the function name ). What I think could be done instead is to add an alias for the `std.objectFields()` as `std.objectKeys()` since the return value (a list of string) doesn't depend on the function name.

PR #1064, opened 2023-04-11
feat: implement std.round

rohitjangid:
Add `std.round` function in standard library.

PR for go-jsonnet: https://github.com/google/go-jsonnet/pull/683

### DIFF
--- a/stdlib/std.jsonnet
+++ b/stdlib/std.jsonnet
@@ -1692,4 +1692,6 @@ limitations under the License.
   xor(x, y):: x!=y,
 
   xnor(x, y):: x==y,
+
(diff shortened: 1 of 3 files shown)


stdlib/std.jsonnet
```diff
@@ -1684,4 +1684,11 @@ limitations under the License.
   sum(arr):: std.foldl(function(a,b)a+b,arr,0),
 
   xor(x, y):: x!=y,
+
+  round(x):: 
+  local arr = std.split(std.toString(x), ".");
```
sparkprime, line 1689: Can't you do `std.floor(x + 0.5)` ?

PR #1067, opened 2023-04-14
Add std.trim for string

Tejesh-Raut:
Add std.trim to standard library.

PR for go-jsonnet: https://github.com/google/go-jsonnet/pull/684

### DIFF
--- a/core/string_utils.cpp
+++ b/core/string_utils.cpp
@@ -44,7 +44,6 @@ UString jsonnet_string_escape(const UString &str, bool single)
             case U'\n': ss << U"\\n"; break;
             case U'\r': ss << U"\\r"; break;
             case U'\t': ss << U"\\t"; break;
-            case U'\v': ss << U"\\v"; break;
             case U'\0': ss << U"\\u0000"; break;
             default: {
                 if (c < 0x20 || (c >= 0x7f && c <= 0x9f)) {
@@ -131,8 +130,6 @@ UString jsonnet_string_unescape(const LocationRange &loc, const UString &s)
                     case 'r': r += '\r'; break;
 
                     case 't': r += '\t'; break;
-
-                    case 'v': r += '\v'; break;
 
                     case 'u': {
                         ++c;  // Consume the 'u'.
--- a/core/string_utils.cpp
+++ b/core/string_utils.cpp
@@ -44,6 +44,7 @@ UString jsonnet_string_escape(const UString &str, bool single)
             case U'\n': ss << U"\\n"; break;
             case U'\r': ss << U"\\r"; break;
             case U'\t': ss << U"\\t"; break;
+            case U'\v': ss << U"\\v"; break;
             case U'\0': ss << U"\\u0000"; break;
             default: {
                 if (c < 0x20 || (c >= 0x7f && c <= 0x9f)) {
@@ -130,6 +131,8 @@ UString jsonnet_string_unescape(const LocationRange &loc, const UString &s)
                     case 'r': r += '\r'; break;
 
                     case 't': r += '\t'; break;
+
+                    case 'v': r += '\v'; break;
 
                     case 'u': {
                         ++c;  // Consume the 'u'.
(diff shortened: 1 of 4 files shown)


stdlib/std.jsonnet
```diff
@@ -1710,4 +1710,6 @@ limitations under the License.
   round(x):: std.floor(x + 0.5),
 
   isEmpty(str):: std.length(str) == 0,
+
+  trim(str):: std.stripChars(str, ' \n\t'),
```
rohitjangid, line 1714: I believe we want to strips `\t` `\n` `\r` `\f` `\v` ` ` in order to make it consistent with golang implementation

doc/_stdlib_gen/stdlib-content.jsonnet
```diff
@@ -458,6 +458,14 @@ local html = import 'html.libsonnet';
             Returns true if the the given string is of zero length.
           |||,
         },
+        {
+          name: 'trim',
+          params: ['str'],
+          availableSince: 'upcoming',
+          description: |||
+            Returns a copy of string after eliminating leading and trailing spaces.
```
rohitjangid, line 466: this needs to be corrected to `trailing whitespaces`

stdlib/std.jsonnet
Tejesh-Raut: updated

core/string_utils.cpp
```diff
@@ -44,6 +44,7 @@ UString jsonnet_string_escape(const UString &str, bool single)
             case U'\n': ss << U"\\n"; break;
             case U'\r': ss << U"\\r"; break;
             case U'\t': ss << U"\\t"; break;
+            case U'\v': ss << U"\\v"; break;
```
sparkprime, line 47: I think there is an issue here because this code is used in the JSON manifester and that will cause it to output invalid JSON because JSON does not have \v.
I think you can simply delete this line and everything will be fine, however, since it will fall back on the \u representation which will be fine.  Come to think of it, the explicit \0 case below this line is probably also unnecessary because the fallback will handle it!

I'm ok with adding \v to the Jsonnet language as long as we don't emit it as JSON.

Can you add \v into the tests in test_suite/unparse.jsonnet?  The golden would have \u000B instead of \v.

Feel free to ping me internally to discuss this (dcunnin@)

core/string_utils.cpp
```diff
@@ -131,6 +132,8 @@ UString jsonnet_string_unescape(const LocationRange &loc, const UString &s)
 
                     case 't': r += '\t'; break;
 
+                    case 'v': r += '\v'; break;
```
sparkprime, line 135: I think this is fine because all it will do is convert "\v" to "\\v" when translating to contexts where the \ is meaningful and we need to escape it.  "\\v" in JSON is fine (it's not necessary but also it's not harmful).  I don't actually even think this is actually even used for JSON strings, it's used for Jsonnet strings (e.g. in the formatter) where you will need to do this.

core/string_utils.cpp
sparkprime: Having said all that, it seems you haven't added \v to the go-jsonnet parser so that means the test you add in this PR will not work in go-jsonnet...

https://github.com/google/go-jsonnet/blob/master/internal/parser/string_util.go#L31

Maybe in that case what you actually want to do is abandon the adding of \v to Jsonnet and instead just use \U000b in your own code :)  It's up to you.

core/string_utils.cpp
sparkprime: On the other hand if we *don't* add \v to Jsonnet then we shouldn't add this either.

core/string_utils.cpp
Tejesh-Raut: I have updated to remove parsing of `\v` (as you mentioned it will cause output to be invalid JSON)

core/string_utils.cpp
Tejesh-Raut: I have updated to remove parsing of `\v` (as you mentioned it will cause output to be invalid JSON)

PR #1071, opened 2023-04-18
feat: implement std.remove and std.removeAt

rohitjangid:
Implement `std.remove` and `std.removeAt`

go-jsonnet PR: https://github.com/google/go-jsonnet/pull/689


doc/_stdlib_gen/stdlib-content.jsonnet
```diff
@@ -1421,6 +1421,26 @@ local html = import 'html.libsonnet';
             |||,
           ]),
         },
+        {
+          name: 'remove',
+          params: ['arr', 'elem'],
+          availableSince: 'upcoming',
+          description: html.paragraphs([
+            |||
+              Remove given <code>elem</code> from <code>arr</code>.
```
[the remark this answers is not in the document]
sparkprime: Actually it seems the go version only removes the 1st one and that's the semantics i'd expect from the name of the function so I guess that's what you wanted.

stdlib/std.jsonnet
```diff
@@ -1710,4 +1710,16 @@ limitations under the License.
...
+
+  remove(arr, elem):: [
+    e,
+    for e in arr
+    if e != elem
+  ],
+
+  removeAt(arr, at):: [
+    arr[i],
+    for i in std.range(0, std.length(arr) - 1)
```
sparkprime, line 1722: I think you can express this as something like:
```jsonnet
local indexes = std.find(arr, at);
if std.length(indexes) == 0 then
  return arr
else
  return arr[0:indexes[0]) + arr[indexes[0] + 1]
```

stdlib/std.jsonnet
```diff
@@ -1710,4 +1710,16 @@ limitations under the License.
   round(x):: std.floor(x + 0.5),
 
   isEmpty(str):: std.length(str) == 0,
+
+  remove(arr, elem):: [
```
sparkprime, line 1714: I think you can express this as something like:
```jsonnet
local indexes = std.find(arr, at);
if std.length(indexes) == 0 then
  return arr
else
  return std.removeAt(arr,indexes[0])
```

So it does not remove all of them

doc/_stdlib_gen/stdlib-content.jsonnet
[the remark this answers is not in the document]
rohitjangid: Yes, it is for first occurrence. Made appropriate changes.

stdlib/std.jsonnet
rohitjangid: Fixed

PR #1078, opened 2023-05-03
feat: implement std.contains

rohitjangid:
Implement `std.contains` function in standard library

go-jsonnet PR: https://github.com/google/go-jsonnet/pull/691


CertainLach:
Isn't it the same thing as std.member?

sparkprime:
Hmm, yes it is.  I had it in my head that member was for sets but the documentation says otherwise!

PR #1079, opened 2023-05-03
feat: implement std.isEqualIgnoreCase

rohitjangid:
Implement `std.isEqualIgnoreCase` in standard library

go-jsonnet PR: https://github.com/google/go-jsonnet/pull/692


sparkprime:
This should be `std.equalsIgnoreCase()` as we already have `std.equals` (and we don't have `std.isEqual`)

rohitjangid:
@sparkprime renamed it to `equalsIgnoreCase`

PR #1082, opened 2023-05-05
feat: std.flattenDeepArray Concatenate an array containing values and arrays into a single flattened array.

bitmaybewise:
## Proposal 

The function `compactArray` recursively flattens multiple arrays, eliminating null values.

Example:

```jsonnet
{
  result: std.compactArray([1, [2, 3], ['4', null], []])
}
```

Outputs:

```json
{
  "result": [1, 2, 3, "4"]
}
```

### DIFF
--- a/doc/_stdlib_gen/stdlib-content.jsonnet
+++ b/doc/_stdlib_gen/stdlib-content.jsonnet
@@ -1271,6 +1271,20 @@ local html = import 'html.libsonnet';
           ],
         },
         {
+          name: 'compactArrays',
+          params: ['arrs'],
+          availableSince: 'upcoming',
+          description: |||
+            Concatenate an array of arrays into a single flattened array, removing null values.
+          |||,
+          examples: [
+            {
+              input: 'std.compactArrays([[1, 2], [], [3, [4]], [[5, 6, [null]], [7, 8]]])',
+              output: std.compactArrays([1, 2, 3, 4, 5, 6, 7, 8]),
+            },
+          ],
+        },
+        {
           name: 'reverse',
           params: ['arrs'],
           availableSince: '0.13.0',
--- a/doc/_stdlib_gen/stdlib-content.jsonnet
+++ b/doc/_stdlib_gen/stdlib-content.jsonnet
@@ -1271,16 +1271,16 @@ local html = import 'html.libsonnet';
           ],
         },
         {
-          name: 'compactArrays',
-          params: ['arrs'],
+          name: 'flattenDeepArray',
+          params: ['value'],
           availableSince: 'upcoming',
           description: |||
-            Concatenate an array of arrays into a single flattened array, removing null values.
+            Concatenate an array containing values and arrays into a single flattened array.
           |||,
           examples: [
             {
-              input: 'std.compactArrays([[1, 2], [], [3, [4]], [[5, 6, [null]], [7, 8]]])',
-              output: std.compactArrays([1, 2, 3, 4, 5, 6, 7, 8]),
+              input: 'std.flattenDeepArray([[1, 2], [], [3, [4]], [[5, 6, [null]], [7, 8]]])',
+              output: std.flattenDeepArray([[1, 2], [], [3, [4]], [[5, 6, [null]], [7, 8]]]),
             },
           ],
         },
--- a/stdlib/std.jsonnet
+++ b/stdlib/std.jsonnet
@@ -865,15 +865,11 @@ limitations under the License.
   flattenArrays(arrs)::
     std.foldl(function(a, b) a + b, arrs, []),
 
-  compactArray(arrs)::
-    local foldable = function(accumulator, value)
-      if std.isArray(value) then
-        accumulator + std.compactArray(value)
-      else if value == null then
-        accumulator
-      else
-        accumulator + [value];
-    std.foldl(foldable, arrs, []),
+  flattenDeepArray(value)::
+    if std.isArray(value) then
+      [y for x in value for y in std.flattenDeepArray(x)]
+    else
+      [value],
 
   manifestIni(ini)::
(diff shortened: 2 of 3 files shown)


sparkprime:
It should give an error in that case.  It's supposed to take an array of arrays and return an array.

bitmaybewise:
> It should give an error in that case. It's supposed to take an array of arrays and return an array.

@sparkprime I believe you're right. It would be misleading to flatten arrays when we only want one level to be flattened.

I've pushed a new function, `compactArray`. I believe it would be better this way. WDYT? Let me know and I will change the description to be a proposal for `compactArray`. 😉

sparkprime:
I think you can write it more simply like this:

```
local old_std = std;
local std = old_std {
  flattenDeep(value):
    if std.type(value) == 'array' then
      [y for x in value for y in std.flattenDeep(x)]
    else if value == null then
      []
    else
      [value],
};

{ 
  simpleVal: std.flattenDeep(0),
  simpleArr: std.flattenDeep([1, 2, 3]),
  mix: std.flattenDeep([1, [2, 3]]),
  mixWithNulls: std.flattenDeep([null, 1, null, [null, 2, null, 3, null]]),
  deeper: std.flattenDeep([[null], [[1], null], [null, 2, [null, [3, null]], null]]),
}
```

But actually I wonder why exclude null by default?  If you remove the part that checks for null and returns `[]` then it will leave the nulls in the linear array.  That might be useful in some cases.  If you don't want them, you can remove them with `[x in std.flattenDeep(foo) if x != null]`.

sparkprime:
BTW it wasn't clear to me if you wanted to support > 2 levels of nesting or whether the way you were recursively calling fold from within the accumulator would actually achieve that.

bitmaybewise:
> ```
> local old_std = std;
> local std = old_std {
>   flattenDeep(value):
>     if std.type(value) == 'array' then
>       [y for x in value for y in std.flattenDeep(x)]
>     else if value == null then
>       []
>     else
>       [value],
> };
> 
> { 
>   simpleVal: std.flattenDeep(0),
>   simpleArr: std.flattenDeep([1, 2, 3]),
>   mix: std.flattenDeep([1, [2, 3]]),
>   mixWithNulls: std.flattenDeep([null, 1, null, [null, 2, null, 3, null]]),
>   deeper: std.flattenDeep([[null], [[1], null], [null, 2, [null, [3, null]], null]]),
> }
> ```

@sparkprime practically speaking, that is basically my proposal, but your implementation has the added benefit that if a single value is fed, it turns it into an array of a single value. It could be useful in some cases where the input could be either null or array(s). 

I also like the approach you mentioned to strip nulls: `[x in std.flattenDeep(foo) if x != null]`. It is an extra step compared to the `compactArray` proposal, but simple and easy too. Array comprehensions tend to become difficult to read and understand, but in this case I don't believe it's harmful.

I usually prefer to represent the absence of values differently, and tend to strip nulls from third party libraries outputs, or transform them before feeding to other functions, but I guess it is a preference of mine rather than a good reason to not keep them.

By simplifying the function to not handle null values makes it simpler (as you said):

```jsonnet
local old_std = std;
local std = old_std {
  flattenDeep(value):
    if std.type(value) == 'array' then
      [y for x in value for y in std.flattenDeep(x)]
    else
      [value],
{ 
  simpleVal: std.flattenDeep(0),
  simpleArr: std.flattenDeep([1, 2, 3]),
  mix: std.flattenDeep([1, [2, 3]]),
  mixWithNulls: std.flattenDeep([null, 1, null, [null, 2, null, 3, null]]),
  deeper: std.flattenDeep([[null], [[1], null], [null, 2, [null, [3, null]], null]]),
}
```

So thinking that not removing nulls would serve more purposes, I believe your suggestion will be more beneficial to the stdlib.

doc/_stdlib_gen/stdlib-content.jsonnet
```diff
@@ -1349,6 +1349,20 @@ local html = import 'html.libsonnet';
...
+          availableSince: 'upcoming',
+          description: |||
+            Concatenate an array containing values and arrays into a single flattened array.
+          |||,
+          examples: [
+            {
+              input: 'std.flattenDeepArray([[1, 2], [], [3, [4]], [[5, 6, [null]], [7, 8]]])',
+              output: std.flattenDeepArray([[1, 2], [], [3, [4]], [[5, 6, [null]], [7, 8]]]),
+            },
+          ],
```
bitmaybewise, line 1364: I did not understand why the input and output are basically the same but the input has enclosed single quotes. As I didn't have the time to investigate, I'm simply replicating the other examples.

doc/_stdlib_gen/stdlib-content.jsonnet
```diff
@@ -1428,6 +1442,8 @@ local html = import 'html.libsonnet';
           description: html.paragraphs([
             |||
               Return the min of all element in <code>arr</code>.
+            |||,
+          ]),
```
bitmaybewise, line 1446: Found this little broken part.

doc/_stdlib_gen/stdlib-content.jsonnet
sparkprime: Yes it's because we don't have a way to convert an expression to a string. Some languages (e.g. C with its macros) can do that.  But we have to have the expression as a string, and then repeat the expression so it can be computed to a value.

sparkprime:
In some ways `[]` already represents `null` in the sense that if you want to return an optional thing deep while building the structure, you can return `[]` instead of `null`.

doc/_stdlib_gen/stdlib-content.jsonnet
bitmaybewise: Ah, got it. Thank you for the explanation.

PR #1083, opened 2023-05-09
feat: Add support for bzlmod

mortenmj:
This PR adds a MODULE.bazel file which allows other repositories to import this one as a bzlmod module.
I don't know how Google handles publishing to the Bazel Central Registry, so I haven't made any attempt at that.

I've added a module extension to grab rules_jsonnet, as it isn't available yet as a module, and to create the python header repository.

All tests are successful when run with `--enable_bzlmod`, and running jsonnet/jsonnetfmt from //cmd succeeds.

I've also made a similar PR for go-jsonnet, which depends on this bzlmod module. That PR can be found here: https://github.com/google/go-jsonnet/pull/698

### DIFF
--- a/MODULE.bazel
+++ b/MODULE.bazel
@@ -1,0 +1,12 @@
+module(name = "jsonnet", version = "0.0.0")
+
+bazel_dep(name = "googletest", version = "1.11.0", repo_name = "com_google_googletest")
+
+register_toolchains("//platform_defs:default_python3_toolchain")
+
+build_defs = use_extension("//tools/build_defs:extensions.bzl", "build_defs")
+use_repo(
+    build_defs,
+    "default_python3_headers",
+    "io_bazel_rules_jsonnet",
+)
--- a/MODULE.bazel
+++ b/MODULE.bazel
@@ -1,12 +1,14 @@
 module(name = "jsonnet", version = "0.0.0")
 
-bazel_dep(name = "googletest", version = "1.11.0", repo_name = "com_google_googletest")
-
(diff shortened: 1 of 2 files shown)


MODULE.bazel
```diff
@@ -0,0 +1,12 @@
+module(name = "jsonnet", version = "0.0.0")
+
+bazel_dep(name = "googletest", version = "1.11.0", repo_name = "com_google_googletest")
```
mortenmj, line 3: Might want `dev_dependency = True` here?

mortenmj:
@sparkprime could you review this PR?

mortenmj:
@sparkprime looks like adding this to the Bazel Central Registry requires that the release process explicitly creates a source archive. The generates ones are not guaranteed to be stable, and cause a test failure when submitting to BCR.

https://buildkite.com/bazel/bcr-presubmit/builds/1361#018857db-81f5-4ff0-b07c-05f5587d41a3

> BcrValidationResult.FAILED: jsonnet@0.20.0 is using an unstable source url: `https://github.com/google/jsonnet/archive/refs/tags/v0.20.0.zip`.
>   | You should use a release archive URL in the format of `https://github.com/<ORGANIZATION>/<REPO>/releases/download/<version>/<name>.tar.gz` to ensure the archive checksum stability.
>   | See https://blog.bazel.build/2023/02/15/github-archive-checksum.html for more context.

Is that something you'd be willing to take a PR to do?
The solutions I've seen to this are either a GitHub Action that does the thing, or various ways of doing it manually.

PR #1085, opened 2023-05-17
feat: Add crypto functions to stdlib

rohitjangid:
Update doc with more crypto functions. Implemented `std.sha1`, `std.sha256`, `std.sha512`, `std.sha3` functions as go only functions.

go-jsonnet PR: https://github.com/google/go-jsonnet/pull/699


rohitjangid:
@sparkprime made the changes to break the compatibility. PTAL.

PR #1087, opened 2023-05-24
feat: implement std.avg 

DeepanshuGoel17:
Add std.avg function in standard library.

PR for go-jsonnet: https://github.com/google/go-jsonnet/pull/700

### DIFF
--- a/stdlib/std.jsonnet
+++ b/stdlib/std.jsonnet
@@ -1703,6 +1703,12 @@ limitations under the License.
 
   sum(arr):: std.foldl(function(a, b) a + b, arr, 0),
 
+  avg(arr)::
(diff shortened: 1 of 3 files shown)


stdlib/std.jsonnet
```diff
@@ -1703,6 +1703,12 @@ limitations under the License.
 
   sum(arr):: std.foldl(function(a, b) a + b, arr, 0),
 
+  avg(arr)::
+    if std.length(arr) == 0 then
+      0
```
sparkprime, line 1708: Don't you want an error on this case like the other ones?

DeepanshuGoel17: Resolved this. Thanks for pointing it out.

PR #1088, opened 2023-05-25
feat: add doc for std.parseCsv and std.manifestCsv

rohitjangid:
Add doc for `std.parseCsv` and `std.manifestCsv`

go-jsonnet PR: https://github.com/google/go-jsonnet/pull/701


rohitjangid:
@sparkprime is it fine to break the compatibility for these functions in cpp implementation?

sparkprime:
Yes I think we can break compatibility for newly introduced library functions at this point.

doc/_stdlib_gen/stdlib-content.jsonnet
```diff
@@ -995,6 +1009,45 @@ local html = import 'html.libsonnet';
...
+            html.pre({}, |||
+              id,name
+              1,foo
+              2,bar
+            |||),
+            html.p({}, |||
+              If <code>json</code> param is not a valid csv compatible object, it would be an error.
+            |||),
+            html.p({}, |||
+              The <code>headers</code> param adds is an optional which would default to all fields in the object.
```
sparkprime, line 1047: It only looks at the first row, though.

doc/_stdlib_gen/stdlib-content.jsonnet
```diff
@@ -681,6 +681,20 @@ local html = import 'html.libsonnet';
             },
           ],
         },
+        {
+          name: 'parseCsv',
+          params: ['str', 'delimiter=","'],
+          availableSince: 'upcoming',
+          description: |||
+            Parses a CSV string into JSON.
```
sparkprime, line 689: Needs to mention how the first row is interpreted as a header row and must have unique column names, which then are used as fields in the JSON representation.

doc/_stdlib_gen/stdlib-content.jsonnet
rohitjangid: Updated

doc/_stdlib_gen/stdlib-content.jsonnet
```diff
@@ -681,6 +681,20 @@ local html = import 'html.libsonnet';
             },
           ],
         },
+        {
+          name: 'parseCsv',
```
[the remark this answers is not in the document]
rohitjangid: Updated

doc/_stdlib_gen/stdlib-content.jsonnet
rohitjangid: Updated

PR #1089, opened 2023-05-26
Add source archive and snippet to release

mortenmj:
@sparkprime ref my comment in #1083, BCR requires that a stable source archive is added to the release. I cribbed this from rules_python, and edited the snippet it generates a bit. Here's an example of what it generates:

https://github.com/mortenmj/jsonnet/releases/tag/v0.21.0

I have no real opinions on what text this release snippet should contain, if anything.

Here's a full example from rules_python: https://github.com/bazelbuild/rules_python/releases/tag/0.22.0

I had to grant GitHub Actions read/write permissions to the repo for this to work. It's in "Workflow permissions" under Settings -> Actions -> General.

### DIFF
--- a/.github/workflows/create_archive_and_notes.sh
+++ b/.github/workflows/create_archive_and_notes.sh
@@ -1,0 +1,53 @@
+#!/usr/bin/env bash
+# Copyright 2023 The Bazel Authors. All rights reserved.
+#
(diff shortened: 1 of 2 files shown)


sparkprime:
Looks like Github Actions already had the permissions.

This doesn't change our release process I guess then.  It will just automatically create this additional artifact and we don't have to worry about it.

PR #1093, opened 2023-06-01
feat: allow negative index/end on std.slice

Duologic:
I found myself implementing this elsewhere and figured it'd be easy to support in stdlib.

I ran `make test`, it succeeded for my changes but had unrelated failures, we might want to fix that in a separate PR.

### DIFF
--- a/stdlib/std.jsonnet
+++ b/stdlib/std.jsonnet
@@ -212,7 +212,7 @@ limitations under the License.
           then 0
           else
             if index < 0
-            then std.length(indexable) + index
+            then std.max(0, std.length(indexable) + index)
             else index,
         end:
           if end == null
--- a/stdlib/std.jsonnet
+++ b/stdlib/std.jsonnet
@@ -208,18 +208,27 @@ limitations under the License.
       {
         indexable: indexable,
         index:
-          if index == null then 0
-          else index,
+          if index == null
+          then 0
+          else
+            if index < 0
+            then std.length(indexable) + index
+            else index,
         end:
-          if end == null then std.length(indexable)
-          else end,
+          if end == null
+          then std.length(indexable)
+          else
+            if end < 0
+            then std.length(indexable) + end
+            else end,
         step:
-          if step == null then 1
+          if step == null
+          then 1
           else step,
         length: std.length(indexable),
         type: std.type(indexable),
       };
-    assert invar.index >= 0 && invar.end >= 0 && invar.step >= 0 : 'got [%s:%s:%s] but negative index, end, and steps are not supported' % [invar.index, invar.end, invar.step];
+    assert invar.step >= 0 : 'got [%s:%s:%s] but negative steps are not supported' % [invar.index, invar.end, invar.step];
     assert step != 0 : 'got %s but step must be greater than 0' % step;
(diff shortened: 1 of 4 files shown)


sparkprime:
@sbarzowski Will this "just work" on go-jsonnet?

I think I was originally opinionated about not supporting this as I wanted the language to be simple... But in retrospect it seems more confusing to support python slicing only "partially" so this seems a good idea.

sbarzowski:
> Will this "just work" on go-jsonnet?

Yes. I think we just desugar to `std.slice`. That said ideally we would add a builtin (but that's neither easier nor harder with this change in). 

BTW this change definitely requires an update to documenation.

johnbartholomew:
It seems like this was almost ready to go back in June, with only a request for documentation updates as a possible blocker. @Duologic would you like to contribute some doc changes for this, or should I work on it?

@sparkprime 
I'm not sure if there is any coordination needed with other jsonnet implementations to get this out? I assume we merge it here first, then go-jsonnet can pick it up (probably with the next release?) and other implementations (e.g., jrsonnet - which has a PR queued already, https://github.com/CertainLach/jrsonnet/pull/117) can update at their leisure.

Duologic:
I just scanned the documentation and didn't find any mention that negative slices are not supported, am I missing something?

Do you want an example added maybe?

johnbartholomew:
Hmm... it's a good point, the tutorial says "Array slices like `arr[10:20:2]` are allowed, like in Python", so perhaps people have always assumed negative slices worked :-)

IMHO adding an example of a negative index in the `std.slice` documentation in the standard library reference.

stdlib/std.jsonnet
```diff
@@ -206,18 +206,27 @@ limitations under the License.
...
         indexable: indexable,
         index:
-          if index == null then 0
-          else index,
+          if index == null
+          then 0
+          else
+            if index < 0
+            then std.length(indexable) + index
+            else index,
```
johnbartholomew, line 214: This can result in `invar.index` being a negative number, which subsequently causes a runtime error as `invar.index` is used as the starting index for iteration, and (currently) array/string indexing does not accept negative values.

Compare Python and Jsonnet (with this patch):
```
$ ./jsonnet -e '("012345")[-6:]'
"012345"
$ ./jsonnet -e '("012345")[-7:]'
RUNTIME ERROR: string bounds error: -1 not within [0, 6)
$ python3 -c 'print(("012345")[-6:])'
012345
$ python3 -c 'print(("012345")[-7:])'
012345
```

johnbartholomew (changes requested):
If you're still willing to work on this PR, could you fix the way `inval.index` is calculated, so that the behaviour of `("012345")[-7:]` matches Python. And add string and array test cases for that in test_suite/slice.sugar.jsonnet. (See my inline code comment)

I realise it's been a while since this PR was first opened, so your motivation or priorities may have changed. If so, please let me know and I'm happy to do the remaining parts and get this merged.

stdlib/std.jsonnet
Duologic: Feel free to work on this, might take a while from me to cycle back to this.

johnbartholomew: Done in 422af989b6d63963e4e6ab693091f544a623857b

johnbartholomew:
Rebased and merged. :-)

PR #1096, opened 2023-06-04
language.html.md - a missing comma

osher:
feels like a syntax error to me...

**EDITED**:

Wait with this. The problem here is that `error` notation is used before it is explained, breaking the reading.
I'll need to propose something different...

**Re-EDITED**:
OK. I finished reading all the doc. There is no explanation of this notation...
does it mean to demonstrate that the expression is not evaluated even as a precompile step?
if so - I suggest to add a clarification about it...

Tell me how to proceed, I'll update the PR

### DIFF
--- a/doc/ref/language.html.md
+++ b/doc/ref/language.html.md
@@ -154,7 +154,7 @@ func(21)
 The arguments are **not** evaluated when a function is called. They are passed lazily and evaluated only when used. This allows expressing things which in other languages require built-in functionality or macros, such as short-circuit boolean operations:
 ```
 local and3(a, b, c) = a && b && c;
-and3(true, false, error "this one is never evaluated")
+and3(true, false, error, "this one is never evaluated")
 ```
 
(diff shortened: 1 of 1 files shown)


sparkprime:
The expression `error e` will expect e to be a string (or convert it to a string) and will then raise an error (which cannot currently be caught) and this error will therefore surface to the user evaluating the config (with a stacktrace which hopefully they find helpful).  It is the fundamental construct that is used to implement assertions and similar things.  It is similar to `throw e` in languages with exceptions.

It seems like there should be a section in the reference manual about it.  There is a section in the tutorial about it.  Maybe the tutorial section can be slimmed down if the full content is in the reference manual.  I have not checked what future examples in the tutorial use in the way of errors, as we don't want examples using constructs that haven't been described earlier in the tutorial.

osher:
aha. I see. First - thanks, that's good to know!

I think that we can add a short explanation about `error` as a part of the demo of the lazy evaluation.

Let me try.

osher:
ok, lmk wdyt. if that's too verbose we can make it a little shorter, e.g.:


```
The expression `error "message"` raises an error with the provided message string. Thus, since `a && b` evaluates to `false` - there is no need to evaluate the value for `c` and so the error is never thrown.
``

sparkprime:
Looks great, thanks

PR #1100, opened 2023-06-27
feat: document std.match in standard library

rohitjangid:
Add documentation for `std.match` function in standard library

go-jsonnet PR: https://github.com/google/go-jsonnet/pull/715


He-Pin:
@johnbartholomew in case you missed this one.

PR #1103, opened 2023-07-17
docs(std.extVar): extvars don't have to be strings, fix the docs.

rudo-thomas:
The docs say that ext vars can only have strings values, which is not true. Both the implementation and the tutorial say otherwise :)

Fix that.


sparkprime:
True, thanks

PR #1108, opened 2023-08-13
Implementing the groupBy

DeepanshuGoel17:
Add std.groupBy function in standard library.


qezz:
Can you please explain how it works? It seems like there are at least several syntax errors

MrG9090 (approved):

PR #1112, opened 2023-11-10
Add MSYS2 packages

podsvirov:
I have created and applied several [patches](https://github.com/msys2/MINGW-packages/pull/19045/files) and I will be glad if they are applied to the original repository.


sparkprime:
Those patches all look reasonable

PR #1120, opened 2024-01-02
add loongarch64 support

212dandan:
Please review. If you have any questions, you can contact me at any time.

### DIFF
--- a/third_party/rapidyaml/rapidyaml/ext/c4core/src/c4/cpu.hpp
+++ b/third_party/rapidyaml/rapidyaml/ext/c4core/src/c4/cpu.hpp
@@ -84,6 +84,11 @@
 #       define C4_CPU_PPC
 #       define C4_WORDSIZE 4
 #   endif
+
+#elif defined(__loongarch64)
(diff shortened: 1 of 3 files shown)


johnbartholomew:
Hi. It appears these are changes to Rapid YAML (the copy of the code in the third_party directory) not to jsonnet itself.

It would be better to contribute these changes upstream to the [Rapid YAML project](https://github.com/biojppm/rapidyaml), and then they can be included in jsonnet when we update the Rapid YAML version. You could also check whether Rapid YAML version 0.5.0 already supports this architecture - if so, then I expect jsonnet will pick it up soon (with PR #1134)

johnbartholomew:
It seems someone has already opened a PR on the c4core project to add support for LoongArch: https://github.com/biojppm/c4core/pull/119

johnbartholomew:
Thanks for your contribution, however I'm going to close this as it is better to get this into Rapid YAML and not try to maintain a modified version of the Rapid YAML code within jsonnet.

PR #1132, opened 2024-02-07
Fix jsonnetfmt --debug-desugaring

johnbartholomew:
The desugaring step handles string backslash escape sequences, but it puts the result back into the same AST elements that the input strings came from. When formatting, these strings need to be re-escaped to turn them into printable string literals.

### DIFF
--- a/core/libjsonnet.cpp
+++ b/core/libjsonnet.cpp
@@ -35,6 +35,8 @@ extern "C" {
 #include "parser.h"
 #include "static_analysis.h"
 #include "vm.h"
+#include "pass.h"
+#include "string_utils.h"
 
 namespace {
 using ::jsonnet::internal::Allocator;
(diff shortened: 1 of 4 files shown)


core/libjsonnet.cpp
```diff
@@ -35,19 +35,48 @@ extern "C" {
...
 using ::jsonnet::internal::RuntimeError;
 using ::jsonnet::internal::StaticError;
 using ::jsonnet::internal::Tokens;
 using ::jsonnet::internal::VmExt;
 using ::jsonnet::internal::VmNativeCallback;
 using ::jsonnet::internal::VmNativeCallbackMap;
+using ::jsonnet::internal::CompilerPass;
+using ::jsonnet::internal::LiteralString;
+
+class ReEscapeStrings : public CompilerPass {
```
sparkprime, line 57: // Used during fmtDebugDesugaring mode to ensure the AST can be pretty-printed

sparkprime (approved):

sparkprime:
Nice

PR #1133, opened 2024-02-09
Reformat everything (all C and C++ code) and enable format checks in GitHub workflow

johnbartholomew:
This touches ~everything by reformatting with clang-format on all .cpp and .h files (by running `make reformat`). And then enables the format checks in the GitHub workflow, to detect regressions.

However, note that in general unless everyone working on the code gets into the habit of having clang-format run automatically before commit (e.g., format-on-save in your editor) then regressions are extremely likely, and that is highly frustrating as it will lead to always failing format checks.

So we should consider whether we want to enforce formatting or not - if not, we can just delete the format-check part of the GitHub workflow and let people style code how they want (rely on manual review to prevent egregious stylistic problems). Or if we do want to enforce it then we probably need a smoother way of doing it - at least good instructions for how to configure various editors to format automatically, and probably a provided git pre-commit hook or something similar.

### DIFF
--- a/.github/workflows/build_and_test.yml
+++ b/.github/workflows/build_and_test.yml
@@ -17,9 +17,6 @@ jobs:
   check_code_format:
     name: Check code formatting
     runs-on: 'ubuntu-20.04'
-    # The code currently doesn't pass formatting.
-    # Enable this workflow job once it does, to detect regressions.
-    if: false
     steps:
       - uses: actions/checkout@v4
       - name: Check clang-format (make test-formatting)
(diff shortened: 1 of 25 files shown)


johnbartholomew:
Currently breaks tests because some of the test data files have a `.cpp` extension which is caught by the auto format command.

Anyway, I'm going to postpone this PR for a while - as @sparkprime points out, there are many PRs in the queue, and if we want to merge any then it's much easier before all the files are touched for reformatting.

Issue #1109, opened 2023-08-19
std.parseYaml fails on non-standard yaml feature, supported in other implementations


CertainLach:
Consider this input: 
```jsonnet
std.parseYaml("0777")
```
cpp-jsonnet outputs:
```
Something went wrong during jsonnet_evaluate_snippet, please report this: [json.exception.parse_error.101] parse error at line 1, column 5: syntax error while parsing array - unexpected number literal; expected ']'
Segmentation fault
```

go-jsonnet outputs:
```
511
```

jrsonnet outputs:
```
511
```

sjsonnet outputs:
```
sjsonnet.Error: Internal error: scala.MatchError: null
Caused by: scala.MatchError: null
```
But this is caused by sjsonnet only supporting objects in parseYaml, so `std.parseYaml("a: 0777")` works:
```
{
   "a": 511
}
```

Why does it happens?
Well, there is two yaml specs: yaml 1.1, and yaml 1.2, and even if it is a minor update by semver... It isn't a semver.
The one breaking change in yaml 1.2, is that octal literals in `0777` form are now forbidden, `0o777` should be used instead.

However, many yaml implementations are not following the spec, especially golang, which will even output octals in `0777` format: https://github.com/go-yaml/yaml/issues/420

For jrsonnet, I had to modify serde-yaml rust library to also accept this input: https://github.com/dtolnay/serde-yaml/pull/225

And rapidyaml did not implemented this compatibility feature: https://github.com/biojppm/rapidyaml/issues/291

johnbartholomew:
I'm upgrading Rapid YAML in #1134, which fixes the segfault. However the behaviour still doesn't match go-jsonnet due to other implementation differences. Specifically, (C++) jsonnet relies on Rapid YAML to emit JSON (ie, to do a direct YAML to JSON conversion) and then parses the output of that just like it would parse any other JSON. But when Rapid YAML converts the input `0777` to JSON it does not emit it as a JSON number, but rather it emits a JSON string literal. This was done deliberately for https://github.com/biojppm/rapidyaml/issues/291

So the 'type' of the value is lost.

We would need to rewrite the `std.parseYaml` builtin to avoid marshalling everything through (Rapid YAML generated) JSON. Or contribute an upstream change to Rapid YAML to make it treat numbers like this differently. Or switch to a different YAML parser.

yuvipanda:
Same thing seems to happen with the following yaml:

```
basehub:
  nfs:
    pv:
[...]

Issue #1050, opened 2023-03-06
std.parseYaml() parses quoted numbers as numbers


vincentmrg:
Hello,

It looks like the `std.parseYaml` function parses quoted numbers as numbers (instead of string).
e.g.
```
bash-5.1$ cat test.jsonnet 
local yaml = 'id: "91372278466"\nname: staging\n';
std.parseYaml(yaml)
bash-5.1$ jsonnet test.jsonnet 
{
   "id": 91372278466,
   "name": "staging"
}
```

Another unexpected behavior is that this function fails if said number starts with a 0.
e.g.
```
bash-5.1$ cat test.jsonnet 
local yaml = 'id: "091372278466"\nname: staging\n';
std.parseYaml(yaml)
bash-5.1$ jsonnet test.jsonnet 
Something went wrong during jsonnet_evaluate_snippet, please report this: [json.exception.parse_error.101] parse error at line 1, column 19: syntax error while parsing object - unexpected number literal; expected '}'
Abort trap: 6
```

sparkprime:
I'm guessing this is an issue with the underlying parser?  Was it C++ or Go

vincentmrg:
It was with the C++ version. I just tested with the go implementation, it works fine.

sparkprime:
It's probably a bug in RapidYAML then

johnbartholomew:
This one should be fixed by #1134 (upgrading RapidYAML to 0.5.0)

johnbartholomew:
Should work on master now. I have added a test case for it (#1136). Thanks for the bug report!

Issue #1014, opened 2022-09-06
std.parseYaml() fails to parse 3 digits versions


ericcitaire:
Simplest way to reproduce :

```bash
$ jsonnet -e "std.parseYaml('version: 1.2.3')"
Something went wrong during jsonnet_evaluate_snippet, please report this: [json.exception.parse_error.101] parse error at line 1, column 16: syntax error while parsing object - invalid literal; last read: '1.2.'; expected '}'
[1]    443873 IOT instruction (core dumped)  jsonnet -e "std.parseYaml('version: 1.2.3')"
```

It seems to be parsing a float instead of a string.

ericcitaire:
This was fixed in rapidyaml a few days ago : https://github.com/biojppm/rapidyaml/pull/299

sbarzowski:
I guess go-jsonnet would work fine, as it uses a different YAML parser. 

We should bump the rapidyaml version here.

jzhn:
Hi, is there any plan to update rapidyaml version in jsonnet to address the issue? Not sure why this issue wasn't fixed in the v0.20.0 cycle.

johnbartholomew:
I'm upgrading Rapid YAML in #1134 - I'll confirm this works then.

johnbartholomew:
Should work on master now. I have added a test case for it (https://github.com/google/jsonnet/pull/1135). Thanks for the bug report!

PR #943, opened 2021-10-19
Add test case for #942 std.parseYaml and block (pipe)

pmorch:
std.parseYaml() doesn't handle this properly:

    f1: |
      a
      b

A test case for issue #942

PR #1134, opened 2024-02-09
Upgrade Rapid YAML to v0.5.0, and switch to the single file amalgamation

johnbartholomew:
Rapid YAML has an extensive CMake based build, which is probably ok but may be awkward to use from our other build systems (e.g., Bazel, which we probably want to continue supporting).

However, Rapid YAML also provides a single-file amalgamation (aka unity build), which is of course much easier to embed.

I expect this makes it harder for users (e.g., distribution packagers) to use a system-provided Rapid YAML install instead. I'd like to fix that, though I'm not sure if anyone is doing that currently - unlike the json lib we don't seem to have prior support for it in the CMake build at least. So it can probably be done as a follow-up. But ideally before the next published release.


johnbartholomew:
A few open bugs should be fixed just by upgrading Rapid YAML:
- #942 (which has a contributed but not-yet-merged test case #943)
- #1014 
- #1050 

YAML-parsing-related bug #1109 is not fixed by the upgrade and needs other work.

PR #1139, opened 2024-03-03
Escape strings printed by `std.assertEqual`

### LINKED ISSUE #363: Strings not escaped in assertEqual errors
Code:
```
std.assertEqual("\n ", "\n")
```

Expected output:
```
RUNTIME ERROR: Assertion failed. "\n " != "\n"
        std.jsonnet:649:13-91   function <anonymous>
        testdata/assert_equal5.input:1:1-28
```

Actual output:
```
RUNTIME ERROR: Assertion failed. 
  != 

        std.jsonnet:649:13-55   function <anonymous>
        testdata/assert_equal5.input:1:1-28
```

Another example where it is pretty important:
```
std.assertEqual("\u001b[31m", "")
```

johnbartholomew:
Fixes #363.

When `std.assertEqual` fails, the error message it produces includes the values that were passed in. This PR uses `std.escapeStringJson` to escape strings passed to `std.assertEqual`. Other values are concatenated as normal (relying on the string concatenation to stringize them as before).

An alternative is to call `std.manifestJson` on both arguments regardless of type. However it's not clear to me that `std.manifestJson` will produce the same output as the (internal) `toString` performed by string concatenation.

PR adds test cases for std.assertEqual with two unequal strings and with two unequal objects.

### DIFF
--- a/stdlib/std.jsonnet
+++ b/stdlib/std.jsonnet
@@ -831,10 +831,14 @@ limitations under the License.
       std.map(map_func, std.filter(filter_func, arr)),
 
   assertEqual(a, b)::
+    // If the values are strings, escape them for printing.
+    // If not, they'll be JSON-stringified anyway by the later string concatenation.
+    local astr = if std.type(a) == 'string' then std.escapeStringJson(a) else a;
+    local bstr = if std.type(b) == 'string' then std.escapeStringJson(b) else b;
     if a == b then
       true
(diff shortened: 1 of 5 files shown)


johnbartholomew:
@sparkprime Do you think this is ok, or would you be concerned about changing the (error-case) output of `std.assertEqual`?

PR #1142, opened 2024-03-09
Rewrite manifestJson to execute "stacklessly" inside evaluate()

johnbartholomew:
This gets rid of the mutual recursion between `manifestJson` and `evaluate`, turning the `manifestJson` processing into a part of the interpreter's main loop. It should remove one avenue for native stack overflows during execution. Of course, the runtime still tracks its own internal call stack depth and can safely abort.

The Interpreter needed to be able to convert things to JSON anyway, as this is the behaviour of string coercion (for string concatenation), and also the behaviour when using the `error` operator/statement.

Some general notes:
- This pokes deeply into the interpreter machinery, which is complex and highly stateful, so there is a high chance of errors in the implementation.
- It's probably _less_ efficient than the previous implementation.
- `Interpreter::evaluate` was already a really long function; this makes it even longer.
- The `Frame` struct gets even bigger.

TL;DR: This is not ready yet, and might never be ready. But hey, it's been fun to make. Consider it just a sketch or prototype for now.


sbarzowski (approved):

sbarzowski:
I think it's a good idea and the code looks reasonable. What is missing?

MrG9090 (approved):

PR #1150, opened 2024-04-11
Document jpathdir option in Python binding docs

danipozo:
Add missing keyword arg to the list.


johnbartholomew:
Thanks for the contribution, and sorry for the delay in processing it.

Issue #1128, opened 2024-02-04
Behavior of `std` overriding and desugaring.


eduardosm:
Go implementation differs from C++ implementation when overriding the `std` binding and invoking an operator that desugars to a `std` library call.

For example with
```jsonnet
local std = {}; 1 == 1
```
which would desugar to
```jsonnet
local std = /* std.jsonnet */; local std = {}; std.equals(1, 1)
```

The C++ implementation fails:
```
RUNTIME ERROR: field does not exist: equals
```
while the Go implementation prints `true`.

To me, the Go behavior makes for sense than the C++ one, since breaking syntax constructs by overriding a binding seems weird. However, the C++ behavior is closer to what the "Desugaring" section of the [specification](https://jsonnet.org/ref/spec.html) mandates.

The Go implementation behaves as if it desugared to something like
```jsonnet
local std = /* std.jsonnet */; local std = {}; (/* std.jsonnet */).equals(1, 1)
```

I think it would be nice to have consistent behavior across implementations, and have such behavior written in the specification.

sparkprime:
The optimisations that replace std functions with their native equivalents need to be aware that std has been redefined, which shouldn't be too hard to implement.

PR #1158, opened 2024-05-09
Bind the standard library to a `$std` variable and use it in desugared expressions

eduardosm:
This keeps redefining `std` from breaking expressions that desugar to calls to standard library functions by redefining `std`.

This is similar to what the Go implementation does.

Fixes https://github.com/google/jsonnet/issues/1128.


MrG9090 (approved):

sbarzowski (approved):

sbarzowski:
Thank you! This makes it cleaner.

MrG9090 (approved):

Issue #1155, opened 2024-04-22
Support underscores or other separators in numeric literals.


jeschkies:
It would be nice to have support for underscores in numeric literals to be able to write `500_000_000` instead of `500000000`. This is a common feature in languages. See e.g. [PEP-0515](https://peps.python.org/pep-0515/).

seizethedave:
This would be valuable. Tons of big numeric constants embedded in jsonnet out there.

This would be a deviation from json, but ECMAScript has digit separators. Would the mods look favorable upon this addition?

sparkprime:
SGTM but there's quite a diversity in approaches across languages. Jsonnet typically tries to be compatible with Python or ECMAScript so what is the difference between PEP-0515 and the ECMAScript approach?

seizethedave:
[ECMA proposal](https://github.com/tc39/proposal-numeric-separator), [PEP515](https://peps.python.org/pep-0515/).

Differences between the two seem to be:
* Python plumbs the support into the standard library, allowing you to do things like `int('123_456')` at runtime. ECMAScript doesn't.

Actually, I'll have to take a closer look later, but among the subset of ECMAScript things that Jsonnet supports, that's the only difference I've noticed so far.

The big one is that both require that an underscore can appear between two numerals (0-9), and nothing else can be touching the underscore. Not an `e`, a `.`, and not another `_`.

seizethedave:
Since that's the only difference I can find, what if we just followed ECMAScript? To put it into words:

* Jsonnet numeric literals should support digit separators.
* We'll follow suit with ECMAScript and Python and use the `_` character.
* Separators may be used in whole numbers.
  * Example: `1_000`, `123_456_789`
* Separators may be used in floating point numbers, both before or after the decimal point.
  * Example: `1_000.123_456`, `3.141_592_653`
* Separators may be used in exponentiated numbers, both before or after the exponent.
  * Example: `1.23e4_567`, `1_000e1_000`
* The implementation won't place any restriction on the size, number, or regularity of groups defined by digit separators. (So `1_1`, `1_1_1_1_1_1`, and `1_11_111_1111` are all valid literals.)
* A `_` may appear between any two numeric digits.
* A `_` may not appear next to any of the other non-digit characters encountered in numeric literals. (Including another `_`.) 
  * Invalid: `1_000_.0`, `1._123`, `7_e3`, `7e_3`, `7e-_3`, `7e+_3`, `1__000`
[...]

PR #1160, opened 2024-06-22
Add digit separators

seizethedave:
This is a draft implementation of adding digit separators (`1_000`) to Jsonnet's numeric constants.

Accompanying issue: https://github.com/google/jsonnet/issues/1155

### DIFF
--- a/core/lexer.cpp
+++ b/core/lexer.cpp
@@ -229,7 +229,8 @@ std::string lex_number(const char *&c, const std::string &filename, const Locati
         AFTER_DIGIT,
         AFTER_E,
         AFTER_EXP_SIGN,
-        AFTER_EXP_DIGIT
+        AFTER_EXP_DIGIT,
+        AFTER_UNDERSCORE
     } state;
 
     std::string r;
@@ -262,6 +263,8 @@ std::string lex_number(const char *&c, const std::string &filename, const Locati
                     case 'e':
                     case 'E': state = AFTER_E; break;
 
+                    case '_': state = AFTER_UNDERSCORE; goto skip_char;
+
                     default: goto end;
                 }
                 break;
@@ -283,6 +286,8 @@ std::string lex_number(const char *&c, const std::string &filename, const Locati
                     case '7':
                     case '8':
                     case '9': state = AFTER_ONE_TO_NINE; break;
+
+                    case '_': state = AFTER_UNDERSCORE; goto skip_char;
 
                     default: goto end;
                 }
@@ -324,6 +329,8 @@ std::string lex_number(const char *&c, const std::string &filename, const Locati
                     case '7':
                     case '8':
                     case '9': state = AFTER_DIGIT; break;
+
+                    //case '_': state = AFTER_UNDERSCORE; goto skip_char;
 
                     default: goto end;
                 }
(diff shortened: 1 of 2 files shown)


sbarzowski:
I'm in favor of this. We need to also:
1) Update the docs. At least https://github.com/google/jsonnet/blob/2bca3a02ac3c06e4dac74c10c4b6c650ae53d148/doc/ref/spec.html#L4. Ideallly also the tutorial. 

See https://github.com/google/jsonnet?tab=readme-ov-file#locally-serving-the-website for working on documentation.

2) Add some end-to-end examples. Just dump some `.jsonnet` files in `test_suite` and run https://github.com/google/jsonnet/blob/master/test_suite/refresh_golden.sh. 

3) We'll need to update go-jsonnet implementation (separate PR). It should be straightforward.

seizethedave:
Thanks @sbarzowski, I'll get cracking on those.

seizethedave:
@sbarzowski I think I'm ready for another round of feedback on this one. I suspect these might need a little more help:
* a more formal treatment of the numeric grammar in the docs
* note how `std.parseInt` and related functions will _not_ honor numbers with underscores
Thanks!

MrG9090 (approved):

MrG9090 (approved):

seizethedave:
Anything I can do to push this forward @sbarzowski? Thanks!

MrG9090 (approved):

MrG9090 (approved):

MrG9090 (approved):

MrG9090 (approved):

MrG9090 (approved):

johnbartholomew:
Sorry for all the delays on this. I do hope to review this soon; it seems like a useful improvement.

MrG9090 (approved):

seizethedave:
@johnbartholomew I'll try to find a few minutes to get this PR back into a healthy/mergable state. If you have any pre-review feedback or changes I can get those going.

core/lexer.cpp
```diff
@@ -262,6 +266,8 @@ std::string lex_number(const char *&c, const std::string &filename, const Locati
                     case 'e':
                     case 'E': state = AFTER_E; break;
 
+                    case '_': state = AFTER_UNDERSCORE; goto skip_char;
```
johnbartholomew: Underscore must not be allowed after a leading zero.

JSON numbers that start with a 0 may only be 0 itself, or a fraction starting `0.`, or a (fairly pointless...) 0 with an exponent (e.g., `0e5`). If you go to the AFTER_UNDERSCORE state after a leading zero then it can be used to produce an invalid number, e.g., `0_5` is read as `05` which is not allowed.

(The reason it's not allowed is because it's ambiguous - Javascript reads numbers with a leading zero as *octal* numbers, but JSON does not support octal numbers, so to avoid a misleading difference between JSON and Javascript leading zeros are simply forbidden)

seizethedave:
> Well, "reviewing this soon" turned into ignoring it for an entire year 😭

Hey, I'm familiar with that operating style.

@johnbartholomew I will take you up on the offer to rebase/fix/submit, thank you very much! Good catches on the state machine.

PR #1167, opened 2024-08-14
Added support for field exemptions for std.prune

raynix:
This is to add support for fields which will be exempted and will not be pruned in an object. For example

```
{
  prune(a, ex=[])::
    local isContent(b) =
      if b == null then
        false
      else if std.isArray(b) then
        std.length(b) > 0
      else if std.isObject(b) then
        std.length(b) > 0
      else
        true;
    if std.isArray(a) then
      [$.prune(x, ex) for x in a if isContent($.prune(x, ex))]
    else if std.isObject(a) then {
      [x]: $.prune(a[x], ex)
      for x in std.objectFields(a)
      if std.member(ex, x) || isContent($.prune(a[x], ex))
    } else
      a,
  containers: {
    local container = {
      image: 'nginx:latest',
      env: [],
      args: [],
      volumeMounts: [
        {
          name: 'test',
          emptyDir: {},
          rubbish: {}
        }
      ],
      resources: {
        requests: {
          cpu: '0.5',
          rubbish: []
        },
        rubbish: [],
      }
    },
    test: $.prune(container, ['emptyDir']),
  },
}
```
the emptyDir will be preserved as a preferred result.


raynix:
Can someone take a look? It's been there for a while :)

PR #1206, opened 2025-01-26
doc/man: Add man pages jsonnet.1 and jsonnetfmt.1

daichifukui:
Hi

With this commit, we'd like to add man pages for the jsonnet and jsonnetfmt commands.
The Debian Lintian quality checker warns that each binary should have a manual page, but hopefully this idea would also apply to non-Debian users.
These man pages are created using the help2man command, so the contents are substantially the same as the output of the help option.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -48,6 +48,7 @@ option(BUILD_TESTS "Build and run jsonnet tests." ON)
 option(BUILD_STATIC_LIBS "Build a static libjsonnet." ON)
 option(BUILD_SHARED_LIBS "Build shared libjsonnet." ON)
 option(BUILD_SHARED_BINARIES "Link binaries to the shared libjsonnet instead of the static one." OFF)
+option(BUILD_MAN_PAGES "Build manpages." OFF)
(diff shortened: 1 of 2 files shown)


johnbartholomew:
It looks like this is the output of the `help2man` tool. What were the exact commands you used to generate this? It's important to be able to re-run this when needed. For example, if command line argumments to jsonnet or jsonnetfmt change, but also whenever there is a new release we'll need to regenerate this since the man page includes the version number.

It would be good to either generate the man pages during the build (and not add them in the repository at all), or at least to have a make target or shell script that can be run to repeatably regenerate the man-pages (and add it to the release checklist)

MrG9090 (approved):

MrG9090 (approved):

daichifukui:
Hi @johnbartholomew , I've updated this change to generate the man pages during the build.
For the exact commands used to generate the man pages, take a look at the files changed.
I'd be happy if you will approve this. Thank you for the review.

johnbartholomew:
Thanks for the contribution, and sorry this was left untouched for so long. Due to #1275 I've rebased it manually to bring this PR up to date.

johnbartholomew:
Merged! Thanks again!

PR #1213, opened 2025-02-24
Allow manually triggering the build-and-test workflow

johnbartholomew:
Occasionally useful to be able to manually trigger the build-and-test workflow (which is normally triggered for pull requests).


johnbartholomew:
Welp; I messed up the merge here, and merged this as part of a different thing. Oops.

PR #1225, opened 2025-03-19
Drop duplicate -O3

jcpunk:
This is already set by `OPT` so no need to specify twice.


johnbartholomew:
Merged. Thanks for the contribution!

PR #1230, opened 2025-04-05
feat: introduce maximum parser recursion depth

thevilledev:
# Problem

Parsing extremely deeply nested structures (e.g., nested arrays [[[...]]], objects {a:{a:{...}}}, or long chains of unary/binary operators) could lead to excessive recursion in the C++ parser functions, potentially exhausting the native call stack and crashing the interpreter.

# Solution

A constant `MAX_PARSER_DEPTH` (set to 1000) is introduced in `core/parser.cpp`. Recursive parsing functions (`parse`, `parseArgs`, `parseParams`, `parseBind`, `parseObjectRemainder`, `parseComprehensionSpecs`, `parseTerminalBracketsOrUnary`, `maybeParseGreedy`, `parseInfix`) now track the current recursion depth and throw a `StaticError` if this limit is exceeded. This ensures the parser terminates gracefully for pathological inputs instead of crashing.

The `current_depth` parameter is generally incremented by 1 whenever a parsing function makes a recursive call to parse a distinct sub-expression or nested structure (e.g., calling `parse` from within `parseObjectRemainder` to handle a field's value, or calling `parse` from `parseInfix` to handle the right-hand operand). This accurately tracks the syntactic nesting depth. Calls that are part of processing the same structural level (like `parseParams` calling `parseArgs` internally) might pass the depth unchanged.

A new test case (`test_suite/error.parse.deep_array_nesting.jsonnet`) has been added to verify that the depth limit is correctly enforced and produces the expected static error. This test case uses deeply nested arrays, which demonstrably causes a segmentation fault on the master branch due to native stack overflow during parsing, but is now caught gracefully by the introduced limit.

I also updated all function doc strings as they were either missing or outdated.

## Why a constant?

Making this limit a constant (rather than a command-line configurable option like `--max-stack`) was a deliberate design choice at this point:

- `MAX_PARSER_DEPTH` guards the parsing stage against C++ stack overflows, while `--max-stack` limits the evaluation stage's Jsonnet VM call stack.
- Exceeding a parse depth of 1000 is highly unlikely for typical, human-written Jsonnet code. It often indicates a structural issue in the code itself rather than a need for a higher limit.
- Adding a configuration option would increase CLI complexity and require plumbing the value through the library. 

That being said, this is open for discussion and feedback.

NOTE: originally reported through Google OSS VRP and public disclosure was preferred.


johnbartholomew:
Looks like a good improvement! It is a little tricky determining what value `MAX_PARSER_DEPTH` should be set to, since the available stack memory is platform-dependent. Unfortunately, avoiding such a hard-coded limit either requires a lot of restructuring or (as you note in the PR description) complexity in plumbing a runtime config value though. I guess 1000 is a reasonable starting point.

I'll try to go through the changes in a little more detail before merging but on a quick skim through this looks good.

core/parser.cpp
```diff
@@ -52,6 +52,12 @@ std::string jsonnet_unparse_number(double v)
     return ss.str();
 }
 
+/** Maximum parsing depth to avoid stack overflow due to pathological or malicious code.
+ * This is especially important when parsing deeply nested structures that could lead to
+ * excessive recursion in the parser functions.
+ */
+static const unsigned MAX_PARSER_DEPTH = 1000;
```
johnbartholomew, line 59: Nit: Move this into the anonymous namespace, to be consistent with the `EMPTY_FODDER` definition.

core/parser.cpp
```diff
@@ -151,12 +157,13 @@ class Parser {
...
-     * \param exprs Expressions added here.
-     * \param end The token that ends the list (e.g. ] or )).
+     * \param args Expressions added here.
      * \param element_kind Used in error messages when a comma was not found.
+     * \param got_comma Whether a trailing comma was found.
+     * \param current_depth Current recursion depth to prevent stack overflow.
      * \returns The last token (the one that matched parameter end).
      */
-    Token parseArgs(ArgParams &args, const std::string &element_kind, bool &got_comma)
+    Token parseArgs(ArgParams &args, const std::string &element_kind, bool &got_comma, unsigned current_depth = 0)
```
johnbartholomew, line 166: IMHO it's better not to provide a default value here. As far as I can tell there are only two call sites and both are changed in this PR to pass a current_depth value. And with a default of 0, forgetting to pass in the depth would be an easy way to introduce a code-path that can cause stack overflow.

The same comment applies for the other functions where you're adding a `current_depth` parameter; best not to give a default value for it.

johnbartholomew (commented):
Looks good :+1: 

Just a couple of minor style comments.

PR #1237, opened 2025-05-09
feat: implement std.isNull

shrred26:
Implement std.isNull in standard library


He-Pin:
@shrred26 What's the usercase of this, to avoid the exception with `if`?

It would be nice to have `?` and `??` in jsonnet.

kalvdans:
What is wrong with using `v == null`?

PR #1240, opened 2025-05-13
Reduce safe integer range to exclude +/- 2^53 and match JavaScript safe-integer bounds

### LINKED ISSUE #1238: BUG: Defined safe integer range is invalid
> Range checks on bitwise operations, and arithmetic shifts are now limited to the integer-safe range (Jsonnet numbers are IEEE double precision floating point, so they can safely represent integers from -2^53 to 2^53).

Jsonnet indeed can represent integers up to 2^53, however "safe" integer is the one that has defined successor and predecessor
2^53 does not have successor, thus arithmetic with it is already invalid, we can't distinguish between 2^53 and 2^53+1, meaning we are not sure if the integer we received from jsonnet code is safe in this case.

For example:

```
// 2^53 + 1 | 0 == 2^53
9007199254740993 | 0 == 9007199254740992
```

The bitwise operation here has silently truncated value, despite documentation stating that bitwise operations perform safe integer range checks.

`9007199254740993` is outside of safe integer range, but safe integer range check does not report error here, as the last bit of this number is already truncated on jsonnet number type side.

In other words: if unsafe integer can pass the safe integer check, then is this safe integer check correct?

Only overflow with larger integer can be detected

```
// 2^53 + 2
9007199254740994 | 0 // STATIC ERROR: numeric value outside safe integer range for bitwise operation.
```

Originally reported at https://github.com/google/jsonnet/pull/1217#issuecomment-2868504070

johnbartholomew:
As noted in https://github.com/google/jsonnet/pull/1217#issuecomment-2868504070 and discussed in #1238.

Although 2^53 is representable (there is a bit-pattern for a IEEE-754 double precision float which represents this value), it is not safe because as @CertainLach pointed out, it is ambiguous/not distinguishable from 2^53+1 which is not representable and would be rounded down.

This also switches from a StaticError to a RuntimeError when encounting a non-integer-safe value since these values are runtime constructs not parse time (static).

@CertainLach if you have time to review, I'd appreciate it.

Fixes #1238.

### DIFF
--- a/core/vm.cpp
+++ b/core/vm.cpp
@@ -873,6 +873,40 @@ class Interpreter {
         }
     }
 
+    /** Safely converts a double to an int64_t, with range and validity checks.
+    *
+    * This function is used primarily for bitwise operations which require integer operands.
+    * It performs two safety checks:
+    * 1. Verifies the value is finite (not NaN or Infinity)
+    * 2. Ensures the value is within the safe integer range [-2^53, 2^53]
+    *
+    * The safe integer range limitation is necessary because IEEE 754 double precision
+    * floating point numbers can only precisely represent integers in the range [-2^53, 2^53].
+    * Beyond this range, precision is lost, which would lead to unpredictable results
+    * in bitwise operations that depend on exact bit patterns.
+    *
+    * \param value The double value to convert
+    * \param loc The location in source code (for error reporting)
+    * \throws RuntimeError if value is not finite or outside the safe integer range
+    * \returns The value converted to int64_t
+    */
+    int64_t safeDoubleToInt64(double value, const internal::LocationRange& loc) {
+        if (std::isnan(value) || std::isinf(value)) {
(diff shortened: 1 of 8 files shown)


CertainLach:
LGTM

(Only users with explicit access to this repository may approve pull requests.)

CertainLach (commented):
LGTM (I misunderstood the test twice initially)

johnbartholomew:
> LGTM (I misunderstood the test twice initially)

That sounds like the tests need some more work. I'll see if I can improve it before merging.

He-Pin:
refs: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/isSafeInteger#description

johnbartholomew:
> That sounds like the tests need some more work. I'll see if I can improve it before merging.

In an event that probably does not surprise anyone, I did not get time to do this so I'm going to merge now, and _then_ work on further improvements :face_with_diagonal_mouth:

Issue #1248, opened 2025-07-16
CMake 4 support


jcpunk:
With the release of CMake 4.0 there are new compatibility requirements: https://cmake.org/cmake/help/latest/manual/cmake-policies.7.html#unsupported-policies

Uncovered compatibility issues via https://fedoraproject.org/wiki/Changes/CMake4.0

PR #1249, opened 2025-07-16
Add cmake4 to cmake_minimum_required.

jcpunk:
Leaving cmake 2.8 may have some long term cost, but for now it all works as expected.

External projects will need to manage their own cmake_minimum_required.

Fixes: https://github.com/google/jsonnet/issues/1248


johnbartholomew:
> Leaving cmake 2.8 may have some long term cost

Given how old CMake 2.8 is, and the significant changes with CMake 3, I would be open to a PR that brings the minimum supported version up to at least 3.5 and possibly something newer (I think 3.5 was released in 2016?). If possible, I would like it to still build on at least the oldest still supported (including LTS support) Debian release (*). At some point the code started to require C++17 so it doesn't seem useful to support a CMake significantly older than the oldest version of GCC that would be able to build it; that might be GCC 9 (https://gcc.gnu.org/gcc-9/changes.html has "The C++17 implementation is no longer experimental." - though experimental C++17 support was available since earlier versions). GCC 9.1 was released in 2019.

(*) Not because Debian is special, but because it's a useful reference point for a likely "old but still in-use version", and it's fairly easy to find what version is available in that distro.

PR #1252, opened 2025-08-27
Mention null field names

kalvdans:
Having a field name expression evaluate to null seems to remove the field from the object. It's a very useful feature so document it in the language spec so that we can rely on the behaviour for any implementation of jsonnet.


johnbartholomew:
Thanks for the contribution!

Issue #1192, opened 2025-01-04
Unable to set default values


laciferin2024:
```
local getExt(varName, default) = 
  if std.extVar(varName) != null then std.extVar(varName) else default;

{
  machine: {
    gpu: 0,
    cpu: 100,
    ram: 100,
  },
  job: {
    APIVersion: "V1beta1",
    Spec: {
      Deal: {
        Concurrency: 1,
      },
      Docker: {
        Entrypoint: [
          "/usr/games/cowsay",
          getExt("Message", "Pass me an input called Message, like darts run cowsay:v1.0.0 -i Message=moo")
        ],
        Image: "ghcr.io/darts2024/cowsay:v1.0.0",
      },
      Engine: "Docker",
      Network: {
        Type: "None",
      },
      PublisherSpec: {
        Type: "local",
      },
      Resources: {
        GPU: "",
      },
      Timeout: 500,
      Verifier: "Noop",
    },
  },
}

```
###Error

```
jsonnet default.dart 
RUNTIME ERROR: Undefined external variable: Message
        default.dart:2:6-25     function <getExt>
        default.dart:19:11-108  thunk from <object <anonymous>>
        Array element 1
        Field "Entrypoint"
        Field "Docker"
        Field "Spec"
        Field "job"
        During manifestation
```

Version: 

```
Jsonnet commandline interpreter (Go implementation) v0.20.0
```

PR #1253, opened 2025-09-03
feat: add extVarWithDefault as builtin to support extVar with a default value

mmsqe:
Updates https://github.com/google/jsonnet/issues/1192


He-Pin:
Nice addition, if this got merged I will implement it in sjsonnet too.

He-Pin:
I think this is a nice addition, then we can override with the extra extVars if provided.

PR #1260, opened 2025-12-01
add loads for cc rules to BUILD files

benjaminp:
builtin rules are removed in newer versions of Bazel


johnbartholomew:
Sorry for the delay in reviewing this. Two notes:

- It looks like `rules_shell` also needs to be added as a dependency in `MODULE.bazel` (it is used to run the test suite)
- Please update the bazel lockfile (you can do this with `bazelisk mod deps --lockfile_mode=update`)

PR #1328, opened 2026-06-26
fix: std.format accepts boolean for numeric conversion codes

He-Pin:
## Summary

The [official Jsonnet spec](https://jsonnet.org/ref/stdlib.html#std-format) states:
> The string formatting follows the same rules as Python.

In Python, `bool` is a subclass of `int`, so `"%d" % True` returns `"1"`. However, jsonnet currently rejects booleans for all numeric conversion codes with `Format required number at N, got boolean`.

This fix coerces boolean values to `0`/`1` at the top of `format_code()` (after the `%s` branch, which correctly passes booleans through `std.toString`), matching Python's behavior.

### Before / After

| Expression | Python | jsonnet (before) | jsonnet (after) |
|---|---|---|---|
| `"%d" % true` | `"1"` | ERROR | `"1"` |
| `"%d" % false` | `"0"` | ERROR | `"0"` |
| `"%f" % true` | `"1.000000"` | ERROR | `"1.000000"` |
| `"%f" % false` | `"0.000000"` | ERROR | `"0.000000"` |
| `"%x" % true` | `"1"` | ERROR | `"1"` |
| `"%o" % true` | `"1"` | ERROR | `"1"` |
| `"%e" % true` | `"1.000000e+00"` | ERROR | `"1.000000e+00"` |
| `"%g" % true` | `"1"` | ERROR | `"1"` |
| `"%c" % true` | `"\x01"` | ERROR | `"\x01"` |
| `"%s" % true` | `"True"` | `"true"` | `"true"` |

## Changes

- **`stdlib/std.jsonnet`**: In `format_code()`, handle `%s` first with the original value, then coerce booleans to `0`/`1` via `local num_val = if std.type(val) == 'boolean' then (if val then 1 else 0) else val;` for all numeric conversion codes (`%d`, `%o`, `%x`, `%f`, `%e`, `%g`, `%c`).
- **`test_suite/format.jsonnet`**: Added assertions for `%d`, `%f`, `%x`, `%o`, `%e`, `%g`, `%s` with boolean values.

## Test plan
- Added assertions for `%d`, `%f`, `%x`, `%o`, `%e`, `%g` with boolean values
- `%s` with boolean continues to produce `"true"` / `"false"` (unchanged)
- Existing tests for numeric and string values remain passing

### DIFF
--- a/stdlib/std.jsonnet
+++ b/stdlib/std.jsonnet
@@ -633,92 +633,96 @@ limitations under the License.
       local zp = if cflags.zero && !cflags.left then fw else 0;
       if code.ctype == 's' then
         std.toString(val)
-      else if code.ctype == 'd' then
-        if std.type(val) != 'number' then
-          error 'Format required number at '
-                + i + ', got ' + std.type(val)
-        else
-          render_int(val <= -1, std.floor(std.abs(val)), zp, iprec, cflags.blank, cflags.plus, 10, '')
-      else if code.ctype == 'o' then
-        if std.type(val) != 'number' then
-          error 'Format required number at '
-                + i + ', got ' + std.type(val)
-        else
-          local zero_prefix = if cflags.alt then '0' else '';
-          render_int(val <= -1, std.floor(std.abs(val)), zp, iprec, cflags.blank, cflags.plus, 8, zero_prefix)
-      else if code.ctype == 'x' then
-        if std.type(val) != 'number' then
-          error 'Format required number at '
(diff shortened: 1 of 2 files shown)


CertainLach:
Given that this change is dubious from correctness standpoint as explained in https://github.com/deltarocks/jrsonnet/pull/233, was rejected in jrsonnet, being rolled-forward in sjsonnet (https://github.com/databricks/sjsonnet/pull/1069), and is not implemented in other impls, this PR should be closed

PR #1336, opened 2026-09-10
fix(stdlib): renormalize %e/%g mantissa carry and handle precision 0

### LINKED ISSUE #1327: std.format %e/%g do not follow documented Python rounding semantics
`std.format` is documented as following Python formatting rules:
https://jsonnet.org/ref/stdlib.html#std-format

In jsonnet v0.22.0, `%e`/`%g` do not renormalize the mantissa when rounding carries from `9...` to `10...`. `%g` precision `0` also appears to be passed through as a negative fractional precision, while Python treats precision `0` for `%g` as precision `1`.

Repro table:

| Expression | Python 3 | C++ jsonnet v0.22.0 | go-jsonnet v0.22.0 | jrsonnet v0.5.0-pre99 |
| --- | --- | --- | --- | --- |
| `"%.0g" % 1.0` | `"1"` | `"5e+00"` | `"5e+00"` | `"0e+00"` |
| `"%+.0g" % 1.0` | `"+1"` | `"+5e+00"` | `"+5e+00"` | `"+0e+00"` |
| `"%.0g" % 0.1` | `"0.1"` | `"5"` | `"5"` | `"0"` |
| `"%.1g" % 9.9` | `"1e+01"` | `"10"` | `"10"` | `"10"` |
| `"%.2g" % 99.9` | `"1e+02"` | `"100"` | `"100"` | `"100"` |
| `"%.0e" % 9.5e10` | `"1e+11"` | `"10e+10"` | `"10e+10"` | `"10e+10"` |
| `"%.1e" % 9.99` | `"1.0e+01"` | `"10.0e+00"` | `"10.0e+00"` | `"10.0e+00"` |

Commands used:

```bash
python3 - <<PY
for expr in [1.0, 0.1, 9.9, 99.9, 9.5e10]:
    pass
print("%.0g" % 1.0)
print("%+.0g" % 1.0)
print("%.0g" % 0.1)
print("%.1g" % 9.9)
print("%.2g" % 99.9)
print("%.0e" % 9.5e10)
print("%.1e" % 9.99)
PY

jsonnet -e '["%.0g" % 1.0, "%+.0g" % 1.0, "%.0g" % 0.1, "%.1g" % 9.9, "%.2g" % 99.9, "%.0e" % 9.5e10, "%.1e" % 9.99]'
```

Likely cause:

In `stdlib/std.jsonnet`, `%g` calls `render_float_dec/sci` with `fpprec - 1`, so `%.0g` passes `-1` precision into the renderer. Separately, the renderer rounds the mantissa but does not normalize `10 * 10^precision` back to `1 * 10^precision` with `exponent + 1`.

This is related in spirit to #879, but that issue was about fixed-point carry for `%f`-style formatting. This issue covers `%e/%g` scientific renormalization and `%g` precision-zero behavior.

jdymitarai:
### Description

This PR fixes two related formatting bugs in `stdlib/std.jsonnet` for `%e` and `%g` conversions (reported in #1327):

1. **Mantissa Carry Renormalization**:
   When formatting floating-point numbers in scientific form (`%e` / `%E` / `%g` / `%G`), rounding to `prec` decimal places can round the mantissa up to `10.0` or higher (for example: `"%.1e" % 9.99` yielded `"10.0e+00"` instead of `"1.0e+01"`, and `"%.0e" % 9.5e10` yielded `"10e+10"` instead of `"1e+11"`).
   When `mantissa * 10^prec + 0.5 >= 10 * 10^prec`, we now detect the carry, increment the exponent by 1, and divide the mantissa by 10 to maintain standard normalized scientific representation (`1 <= mantissa < 10`).

2. **`%g` Precision of 0 & Boundary Transition**:
   Per Python format specifications (which Jsonnet follows), a precision of 0 for `%g` is treated as equivalent to a precision of 1. Previously, precision 0 resulted in passing negative precision to `render_float_dec`, leading to corrupted output such as `"%.0g" % 1.0` -> `"5e+00"`.
   In addition, when rounding produces an exponent carry (e.g. `"%.1g" % 9.9`), the effective exponent is updated to `1`, correctly switching to scientific notation `"1e+01"` instead of decimal overflow.

Fixes #1327.

### Verification
- Added automated regression tests to `test_suite/format.jsonnet` for `%e`, `%E`, `%g`, and `%G` testing mantissa carries, precision 0, and boundary exponent transitions.
- Verified against Python standard library formatting across comprehensive test cases (`test_suite/format.jsonnet` returns `true`).
- C++ test suite passes without regressions.


jdymitarai:
Closing this PR to avoid review overhead on the team. The context remains in the thread for future reference if helpful. Thanks for your time.

